PR #6, opened 2026-06-23
Populate radio dropdowns instantly from a prebuilt catalog + add search

lanrat:
First-load latency was dominated by list_registered_radios() importing all 191 CHIRP drivers in Pyodide, each triggering a blocking per-file jsDelivr fetch via the import hook, so the make/model dropdowns could not render until that whole sweep (plus Pyodide cold start) finished.

- Add scripts/build-catalog.mjs (npm run build:catalog, also run by build:dist) which precomputes vendor/model/module/class/baud/isLive into web/radio-catalog.json from the local CHIRP submodule.
- handleListRadios now fetches the static catalog first and only falls back to live in-browser enumeration if it is missing/unusable; Pyodide boots lazily on first selection/CSV/serial action instead of blocking the list render.
- Add a search box above the make/model dropdowns that filters radios by make or model (multi-token, case-insensitive) across all vendors, preserving the current selection and clearing automatically for programmatic selections.
- Add UI test coverage for the search filtering; record findings/release notes.


web/js/runtime-rpc.js
```diff
@@ -91,6 +117,20 @@ async function loadRadioCatalogFromSources() {
   return radioCatalogCache;
 }
 
+// Resolve the radio catalog, preferring the prebuilt static file so the
+// dropdowns appear without waiting on Pyodide + per-driver imports.
+async function loadRadioCatalog() {
+  if (radioCatalogCache) {
+    return radioCatalogCache;
+  }
+  const fromStatic = await loadRadioCatalogFromStatic();
```
jasiek, line 142: The committed catalog reports CHIRP revision `61a03fc…`, but this runtime is pinned to `1467519…`. Because the loader accepts the static catalog without validating `data.chirpRevision`, the UI can expose driver classes absent from the runtime or omit classes that the runtime supports. Please generate the catalog from the runtime-pinned revision or reject a mismatched catalog and fall back to live enumeration.

web/js/ui.js
```diff
@@ -2035,18 +2090,15 @@ export function createUiController() {
       }
     });
 
+    radioSearchEl?.addEventListener("input", () => {
+      radioFilterText = String(radioSearchEl.value || "").trim();
+      refreshMakeOptions();
+      reloadForSelectedRadio();
```
jasiek: This starts metadata and settings loads on every search keystroke. Rapid input can leave several requests in flight, and an older response can overwrite `radioMetadata`/`radioSettingsState` after `selectedRadio` has changed. It also boots Pyodide merely from filtering. Please avoid loading until an explicit radio selection, or add cancellation/request-version guards before applying results.

jasiek (commented):
Two correctness issues need addressing before merge.

web/js/runtime-rpc.js
lanrat: Fixed.
 The CHIRP revision now lives in a single shared constant (DEFAULT_CHIRP_REVISION in python-sources.mjs, set to the submodule/catalog revision the Node test suite validates against), the runtime rejects a static catalog whose chirpRevision doesn't match it exactly and falls back to live enumeration, and build:catalog refuses to emit a catalog when the submodule disagrees with the pin.

web/js/ui.js
lanrat: Fixed.
Search keystrokes no longer start loads directly: the load is debounced until typing settles and skipped when the effective selection didn't change, so filtering alone no longer boots Pyodide. Every metadata/settings load also carries a request-version token and stale responses are discarded. Separately, all Pyodide-backed RPCs are now serialized through a call queue, which also fixes a pre-existing duplicate-driver-registration race you hit while testing #12 (details there).

jasiek:
This has been merged in, with some changes, but github didn't pick it up. I'll close it to avoid confusion.

PR #7, opened 2026-07-02
Feature/channel copy paste

lanrat:
Add support for copy/paste for channels.


web/js/ui.js
```diff
@@ -1258,6 +1270,200 @@ export function createUiController() {
...
+    }
+    try {
+      await navigator.clipboard.writeText(payload.tsv);
+    } catch (error) {
+      logDebug(`CLIPBOARD write failed: ${error}`);
+      setStatus(`Clipboard write blocked; press Ctrl+${remove ? "X" : "C"} / Cmd+${remove ? "X" : "C"} instead.`);
+      return;
+    }
+    if (remove) {
+      removeSelectedChannelRows();
```
jasiek: `payload` snapshots the clipboard text and count, but after awaiting `navigator.clipboard.writeText()` this removes whatever rows are selected at completion. If a permission prompt delays the write and the selection changes, Cut deletes different rows than it copied. Please include the original indexes in the payload and delete those exact rows after a successful write.

web/js/clipboard.js
```diff
@@ -0,0 +1,205 @@
...
+    return null;
+  }
+  const records = parseTsv(text).filter((record) => !recordIsBlank(record));
+  if (records.length === 0) {
+    return { rows: [], usedHeader: false };
+  }
+  const candidateHeaders = records[0].map(
+    (field) => HEADER_BY_LOWERCASE.get(field.trim().toLowerCase()) ?? field.trim(),
+  );
+  const usedHeader =
```
jasiek, line 147: Requiring both `Location` and `Frequency` misclassifies valid spreadsheet header subsets. For example, `Name<TAB>Frequency` becomes two positional data rows: the header itself is imported as a channel and subsequent values shift into the wrong columns. Please recognize a row as a header based on known column names without requiring `Location`, and add coverage for subsets that omit it.

jasiek (commented):
Two correctness issues should be addressed before merge.

web/js/ui.js
lanrat: Fixed. The cut payload now captures the selected row objects, and a successful clipboard write removes exactly those rows (identity-based, shared with the normal remove path). This stays correct even if the selection or row order changes while the write is parked behind a permission prompt, which capturing indexes wouldn't.

web/js/clipboard.js
lanrat: Fixed. A first record now counts as a header when at least two of its fields are known CHIRP column names and the known names aren't outnumbered by unknown ones, so any subset works (Name\tFrequency, Name\tTone\tPower, …) without requiring Location or Frequency, the header row is no longer imported as a channel, and columns don't shift. Extra unknown columns are still tolerated, and a data row whose single cell happens to match a column name (e.g. a channel named "Tone") stays positional. Added tests for the subset cases and both misclassification guards; the subset tests fail on the previous code.

PR #8, opened 2026-07-03
Add WebUSB serial support: FTDI + Prolific PL2303 drivers for Android Chrome

lanrat:
Chrome on Android (148+) exposes `navigator.serial`, but it only drives a limited device set — not the FTDI/PL2303-class USB-UART chips used by most radio programming cables. This PR adds a WebUSB-based serial path with native chip drivers, verified end-to-end on Android Chrome with real radios.

## What's included

- **FTDI driver** (`web/js/ftdi-webusb.js`): FT231X/FT232R etc., libftdi-faithful init (baud divisor math, FIFO purges, latency timer, 2-byte status-header stripping). Verified with an FT231X cable + Baofeng UV-5R (connect + clone download).
- **Prolific PL2303 driver** (`web/js/pl2303-webusb.js`): detects the chip generation from the device descriptor (01 / HX / TA / TB / newer HXN: GC/GB/GT/GL/GE/GS) and applies the matching init + register map, with DTR/RTS control-line support. Verified on Android with a PL2303 cable + real radio.
- **CDC-ACM fallback**: devices that are neither FTDI nor Prolific dispatch to Google's `web-serial-polyfill` (lazily imported only when needed).

## How it works

- A single WebUSB device chooser (filters: FTDI vendor id, Prolific vendor id, CDC class codes) dispatches the chosen device to a chip-specific driver exposing the Web Serial `SerialPort` surface (`open`/`readable`/`writable`/`setSignals`/`getInfo`/`close`), so `BrowserSerialBridge` works identically over native Web Serial and WebUSB.
- Native Web Serial is always preferred. A **Connect via WebUSB** button forces the WebUSB transport — needed on Android where native Web Serial exists but cannot drive these chips, and there is no API to detect that in advance.
- Protocol references: the Linux kernel drivers (`ftdi_sio.c`, `pl2303.c`) and usb-serial-for-android; the PL2303 work was cross-checked against the MIT-licensed WebUSB ports descended from andreasgal/usbserial and fixes two request-type bugs found in them.

## Notable fix

The FTDI read path originally deadlocked: a `ReadableStream` `pull()` that resolves without enqueuing (status-only packet from an idle chip) is never re-invoked per the Streams spec, so all reads wedged after the first idle packet. Both drivers poll inside `pull()` until real payload is enqueued, and recover stalled bulk IN endpoints with `clearHalt`.

## Testing

- `npm run test:serial` — 22 unit tests: transport selection/fallback, FTDI baud-divisor vectors and init sequence, status-byte stripping, the read-deadlock regression, PL2303 chip-detection ladder, HX vs HXN init sequences, line-coding bytes, DTR/RTS state transitions, chooser dispatch.
- Hardware-verified on Android Chrome: FTDI and PL2303 cables, radio clone download completing. Desktop native Web Serial behavior unchanged.
- Not supported (documented): CH340/CP2102 — they need chip-specific drivers; the two driver files show the pattern to add more.


jasiek:
Thank you for your contributions. This looks like a useful change - what would it take to support the CH340/CP2102 - do you have a cable with these chips? If not, I'll see if I can dig them out and augment this.

lanrat:
It should be possible to add support for CH340/CP2102, or any other usb-serial chip, but I don't currently have any radio's that use that chip to test with.

This PR lays the groundwork for any usb-serial chips. I would suggest merging this PR, and once merged, adding support for other usb-serial chips should be pretty straightforward. I'd even be happy to draft a future PR for them, but I would not be able to test without the actually hardware.

PR #9, opened 2026-07-03
Fix unreadable, overlapping channel-table headers on desktop

lanrat:
table-layout: fixed with width: 100% compressed all 17 columns to fit the window, so on typical desktop widths each header got ~1/17 of the space and its text spilled over the neighboring column (e.g. "FrequenDuplex"). The mobile pass already solved this below 900px by letting the table take its natural width with no-wrap headers and scrolling horizontally inside .mem-table-scroll; apply the same approach at all viewport sizes. The media query keeps only the mobile-specific taller touch targets.

example of broken table headers:

image


jasiek:
I think I'll merge this in - what do you think about using a completely different widget in place of a row for editing these on a mobile device?

lanrat:
I have no objections.

PR #10, opened 2026-07-03
Add clone progress bar for radio download/upload

lanrat:
CHIRP drivers already report per-block progress via radio.status_fn; _status_to_log flattened it into one debug log line per block. It now forwards each report to the UI through a new serial_progress JS global (same pattern as serial_log): runtime-rpc registers the global, the serial RPC handler gains a "progress" op with an onProgress sink, and app.js wires it to ui.updateCloneProgress.

The UI is a slim native <progress> bar in the Serial Bridge panel under the Download/Upload buttons, hidden when idle: it appears indeterminate when a clone starts, then shows the driver's own phase message plus a percentage once block counts arrive (drivers reporting no counts keep the indeterminate state; -1/-1 is the sentinel). Both clone handlers hide it in a finally block.

Debug log noise drops as a side effect: status lines are logged only on phase-message changes instead of once per block. Progress display failures are swallowed in Python so they can never break a clone.
image


web/js/ui.js
```diff
@@ -155,6 +159,49 @@ export function createUiController() {
...
+  // when a driver reports no counts; the bar then stays indeterminate).
+  function updateCloneProgress(cur, max, msg) {
+    if (!cloneProgressEl) {
+      return;
+    }
+    cloneProgressEl.hidden = false;
+    if (msg && cloneProgressLabelEl) {
+      cloneProgressLabelEl.textContent = msg;
+    }
+    if (Number.isFinite(cur) && Number.isFinite(max) && max > 0 && cur >= 0) {
```
jasiek, line 188: When a driver switches from a counted phase to a no-count phase (`-1/-1`), this condition is false, so the `<progress>` element keeps its previous `value` and the old percentage text remains visible. The label changes, but the UI still shows a stale determinate percentage, contrary to the sentinel behavior described in this PR. Please handle the fallback by removing `value` and clearing the percentage, and cover the determinate-to-indeterminate transition in a UI-level test.

web/python/runtime_bridge.py
```diff
@@ -508,15 +509,30 @@ def _fallback_constructor():
     return radio
```
jasiek, line 509: This deduplication state persists across clone operations. If a driver reports the same message throughout a transfer, the next transfer can suppress that phase message entirely because `_last_status_msg` still contains it. Please reset the state at the start of each clone, or keep it in a per-radio/per-operation callback closure.

jasiek (commented):
Two review findings:

web/js/ui.js
lanrat: Fixed. No-count reports now remove the value attribute (making the bar indeterminate, same as beginCloneProgress) and clear the percentage text, so a -1/-1 phase can't display the previous phase's percentage. Added the UI-level test you suggested: it drives determinate → indeterminate → determinate through the controller and fails against the previous code.

web/python/runtime_bridge.py
lanrat: Fixed with the per-operation closure you suggested. _make_status_logger() holds the dedup state and each radio.status_fn assignment gets a fresh one, so a driver that repeats one message through a whole transfer can't suppress that message from the next transfer's log. Verified in the runtime: two consecutive operations with an identical repeated message each log it exactly once.

PR #12, opened 2026-07-03
Ask before CSV import destroys existing channels: replace, merge, or cancel

lanrat:
Importing a CSV silently replaced the whole channel list, so an accidental import discarded any unsaved edits with no warning. The csv-file handler now parses the file first (side-effect free), and if the editor holds real channels (usable frequency or name - blank inserted rows do not count) shows a modal stating how many channels would be lost and how many the file contains, with three choices: Replace (previous behavior), Merge (append imported channels below the existing ones and renumber Locations), or Cancel. Escape and clicking outside the dialog cancel. Imports into an empty editor and the sample loader are unchanged.

loadCsvText is split into parseCsvViaRuntime + applyParsedCsv(mode) so the prompt can sit between parsing and applying; the modal reuses the existing modal-overlay/modal-card styling and is promise-based (askImportChoice resolves "replace" | "merge" | "cancel").


jasiek:
LG but I need to manually test

lanrat:
I have all of my PRs merged and running on github pages for my fork if it helps for your testing.

https://lanrat.github.io/webchirp/

jasiek:
When I go to this site and select any brand, I get this message:

```
...
  File "/webchirp_runtime/chirp/directory.py", line 72, in register
    raise Exception("Duplicate radio driver id `%s'" % ident)
Exception: Duplicate radio driver id `HobbyPCB_RS-UV3' (or another model)
```

lanrat:
Good catch.  I tracked this down and it's not related to this PR (or the demo site's merged branches). It's a pre-existing race that reproduces on master: selecting a radio fires the metadata and settings RPCs concurrently, both import the selected driver module, and the lazy import hook suspends mid-import (JSPI run_sync) while fetching driver source from the CDN. The second import then re-executes the module body before the first finishes, and directory.register raises Duplicate radio driver id for whichever driver the brand resolves to (hence "or another model"). It's timing-dependent, which is why it didn't show up in my earlier testing.

I've fixed it in #6 (serializing all Pyodide-backed runtime calls through a FIFO queue), along with the two review findings there. The demo site at https://lanrat.github.io/webchirp/ is updated with the fix, so it should be usable for testing this PR now.

PR #20, opened 2026-07-27
Report what actually happens in the app, not just that buttons were pressed

jasiek:
Analytics covered two events, `radio_download` and `radio_upload`, both fired *before* the operation ran. So the reports counted button presses: a driver that fails on every radio it is pointed at looked exactly like one that works, and everything outside the clone path — the boot, the browser capability check, file import/export, repeater queries, the settings editor — was invisible.

### 352f9e7 — Report how clones end, not just that they were attempted

Every clone now reports its outcome alongside the existing attempt event: success with duration and channel count, failure with duration, the stage it failed at, and the cause. Together that gives a per-driver success rate — the closest thing to a browser compatibility matrix we can collect, and data nobody else has, since desktop CHIRP doesn't phone home.

Uploads additionally report when CHIRP's own preflight blocked them and which column it rejected first, so a field users cannot get past shows up as a pattern rather than as scattered bug reports.

Error text is mapped onto a fixed vocabulary before it is sent (`timeout`, `no_response`, `ident_mismatch`, `checksum`, …), because GA4 drops high-cardinality parameters and raw messages carry radio contents. The exception *type* name rides alongside, so an unanticipated failure still reports as `RadioError` rather than collapsing into `other` with nothing to go on.

### 87cf7ab — Measure the funnel from page load to a connected radio

A visitor who left because their browser has no Web Serial, because the boot took too long, or because the port would not open was indistinguishable from one who never showed up. Four steps now report:

- `app_ready` carries the cold-start duration, radio count, and which catalog path filled the dropdowns. The static-catalog fallback is silent today and costs a full Pyodide driver import before anything appears.
- Browser serial capability rides as a *parameter* on `app_ready`/`app_init_failed` rather than as its own event, so the share of visitors who can never reach a radio shares a denominator with everyone who got that far.
- `radio_selected` reports the radio, how it was chosen (search / dropdown / restored cookie / detected from a loaded image), and whether it is a live-mode radio — a dead end in this UI.
- `serial_connected` reports the transport that won next to the one that was asked for; `serial_connect_failed` separates a closed port picker from an adapter the browser could not open.

### 6db020e — Measure which of the features we built get used

`codeplug_import` / `codeplug_export`, `repeater_import` (with the **result count** — a query returning nothing means the filters or the proxy are wrong, which is invisible today), `repeater_modal_opened` to give it a denominator, `repeater_geolocate`, `preset_channels_added`, `channel_edit`, `settings_view_opened`, `settings_tab_opened`, `report_issue_clicked`, `debug_log_copied`.

`channel_edit` reports each column once per session, so it answers which fields people use without one event per keystroke.

### b3b86bc — Report the scale and provenance of what people edit

Added as dimensions on the events above rather than as new events, so nothing is counted twice. `channel_count_bucket` turns codeplug size into something every report can group by. `codeplug_source` is the part not derivable from anything else: whether the rows in the editor came off a radio or out of a CSV or `.img` — and `mixed` once anything has been merged in. On the write path that turns `radio_upload_success` into an answer to whether people upload what they just read back, or a file they brought with them.

### d2b662d — Report analytics only from the production deployment

Every copy of this app carries the same measurement ID — dev servers on localhost, forks serving their own Pages site, anyone who clones it — and all of them reported into the same GA4 property. One developer reloading localhost all day was indistinguishable from real traffic. The events above make that worse rather than better: a per-driver success rate is worthless if the failures are someone's laptop with no radio plugged in.

The **tag** is gated rather than only our own events — the vendor snippet sends `page_view` and `session_start` by itself, and that is the bulk of what has to stay out. `www` is allowed beside the apex because `CNAME` points at the apex and an unredirected `www` visitor is real traffic; forks are unaffected, being on `github.io` or their own domain.

---

## Reconciliation with master (f9ee2cb, afb2807, 19d3d4e, 25d8899)

Master grew analytics of its own while this sat: `web/js/analytics.js` shared by both pages, a `CUSTOM_DIMENSIONS` registry, `npm run ga:dimensions` to sync the GA4 property, and `scripts/test-ga-dimensions.mjs` which fails the build on an undeclared parameter. This branch now builds on that rather than beside it.

- **One gtag, one import path.** `trackEvent` lives in `web/js/analytics.js`; `web/js/ui/analytics.js` keeps the UI-side parameter helpers and re-exports it. It now also swallows a throwing `gtag` — content blockers commonly replace it with a stub that throws.
- **The host gate moved into the module.** Neither page carries a `googletagmanager` tag at all; the module injects the loader past the hostname check, so off the production host nothing is requested from the vendor rather than only our events being suppressed. Verified in headless Chrome from both sides: on `127.0.0.1`, `gtag` is `undefined`, `dataLayer` absent and zero requests to Google; with `codeplug.org` resolved to the dev server, the loader is requested exactly once and `config` is sent once.
- **`<meta charset>` is back at the top of both heads**, inside the 1024 bytes the HTML spec requires (review finding 2). A test pins the byte offset.
- **All 30 parameters are declared as custom dimensions**, which is what the original description asked for and could not do. Two renames on the way, because the names collided across event sites: `source` → `repeater_source` / `import_source`, and `mode` → `import_mode`.
- **`sample_loaded` is gone** — master removed the sample codeplug, so the event and its handler went with it.
- **RSGB ETCC is instrumented too.** It landed on master after this branch was cut, so the newest repeater-import path was the one reporting nothing; it now sends the same three events under `repeater_source: "rsgb"`.

### Review findings

1. **Merge conflict / rebase** — resolved; `package.json` keeps both test lines.
2. **`<meta charset>` past 1024 bytes** — fixed and pinned by a test.
3. **`codeplug_source` never became `mixed`** — fixed at `insertRowsAtSelectionOrEnd`, the single choke point for repeater queries, RSGB queries and band-plan presets. Load off a radio, add 200 repeaters, upload: now reports `mixed`, not `radio`.
4. **Attempt-vs-outcome** — `sample_loaded` is moot; imports gained the missing failure counterpart, `codeplug_import_failed`, carrying the format, how the file arrived and the classified cause.
5. **`radio_selected` double-counts via the `make` path** — documented at the call site: both events are real, and `method` separates the radio people chose from the one they passed through.
- **Copilot, mid-flight selection change** — `downloadFromRadio`/`uploadToRadio` now use the captured `radio` for their status text and their runtime payload, not `state.selectedRadio`.
- **Copilot, `var gtagScript` leaking a global** — moot; the injection moved into the module.

## What reviewers should know

**One behavioural change outside analytics.** `listRadios` returns a `source` field (`"static"` | `"sources"`) alongside `radios` — additive, no consumer breaks. It is what `catalog_source` reports, and the fallback it exposes is otherwise silent.

**One proposed event was dropped as unreachable.** `settings_view_blocked` would have read a permanent zero: `updateViewButtons()` disables the Radio Settings tab on exactly the condition the click handler guards on. The reason is a comment at `web/js/ui.js`.

**What is never sent.** No file name, channel name or comment, frequency, setting value, search term, or coordinate. Every value is a CHIRP driver identifier, a CHIRP column or settings-group name, a fixed enum, or a bucketed count. Geolocation reports only that it was used and whether it worked — the coordinates stay in the form. The rules are written at the top of `web/js/ui/analytics.js` and in `AGENTS.md` so the next event site inherits them.

**The GA property needs syncing before merge.** `npm run ga:dimensions` plans 34 creates and 2 description updates — the 30 parameters here plus 4 master declared but never applied; the property currently holds 2. GA4 does not backfill, so a parameter registered after merge is invisible for everything collected before it. That is 36 of the 50 event-scoped slots.

## New dependencies

None.

## Testing

- `npm test` — 209 + 32 + 553 + 11 pass. `scripts/test-analytics.mjs` covers the module, the host gate and both pages; `scripts/test-analytics-events.mjs` covers the parameter helpers; `scripts/test-ga-dimensions.mjs` fails on any parameter sent but not declared.
- Driven in headless Chrome against the dev server, on `127.0.0.1` and with `codeplug.org` mapped to it, with googletagmanager blocked at the network layer throughout so the production-host run could not reach GA. App boots clean either way, no console errors.


jasiek:
## Review — Approve with nits, needs rebase ⚠️

The design is sound and carefully executed: all `gtag` access is centralized in `web/js/ui/analytics.js` behind a guard and try/catch so telemetry can never throw into the main flow; outcome events fire after outcomes with data captured before the awaits; payloads contain no PII (verified at every event site — driver identifiers, fixed enums, column/group names, bucketed counts only); and the UI module conventions are respected. `scripts/test-analytics.mjs` genuinely exercises the host gate by eval'ing the real bootstrap from `index.html`.

**Findings:**

1. **Major — needs rebase, merge conflict in `package.json`**: master added `scripts/test-ch340-webusb.mjs` to the `test:channels` line while this branch added `scripts/test-analytics.mjs` to the same line (confirmed via `git merge-tree`). Resolution is trivial — keep both. No other overlap: master's commits since branching touch only files this PR doesn't modify, and the PR already targets the modularized `web/js/ui/` layout.
2. **Minor — `<meta charset>` pushed past the 1024-byte limit** (`web/index.html`): the new 20-line comment moves `charset="UTF-8"` from byte 338 to byte 1607, violating the HTML spec's requirement that the encoding declaration sit within the first 1024 bytes — and the comment contains em dashes (non-ASCII) *before* the declaration. Harmless when the server sends `charset=utf-8` (GitHub Pages does), but risks mojibake on `file://` or a bare dev server. Fix while rebasing: shorten the comment or move `<meta charset>` above it.
3. **Minor — `codeplug_source` never becomes `"mixed"` for repeater imports or presets**: `web/js/ui/state.js:12-15` documents `"mixed"` once an import is merged, but only `applyParsedCsv` in `codeplug-io.js` sets it; `repeater-query.js` and band-plan preset insertion go through `insertRowsAtSelectionOrEnd` without touching it. Load sample → import 200 repeaters → upload reports `radio_upload_success` with `codeplug_source: "sample"` — mislabeled provenance on exactly the write-path question this dimension exists to answer.
4. **Nit — `sample_loaded` fires before the outcome is known** (`codeplug-io.js` bindEvents): the event fires before `runFileLoad(...)`, so a failed sample load still counts — the attempt-vs-outcome pattern the PR itself argues against. Relatedly, `codeplug_import` has no failure counterpart.
5. **Nit — `radio_selected` double-counts via the `"make"` path** (`web/js/ui/radio-catalog.js:439-441`): changing the vendor fires `radio_selected` for the auto-defaulted first model, then the real pick fires a second event. The `method` dimension lets analysts filter, but worth documenting.

PR #26, opened 2026-07-28
Align endpoint handling across the WebUSB serial drivers

### LINKED ISSUE #25: Align endpoint handling across the WebUSB serial drivers
Follow-up from the CH340 driver review in #24. The three WebUSB drivers select and validate their bulk endpoints differently, and two of them have gaps that only stay harmless because of how current hardware happens to enumerate.

## PL2303: hard-coded interface number in the failure message

[`web/js/pl2303-webusb.js:288`](https://github.com/jasiek/webchirp/blob/master/web/js/pl2303-webusb.js#L288)

```js
throw new Error("PL2303: bulk IN/OUT endpoints not found on interface 0");
```

The interface is picked by array index (`configuration.interfaces[0]`), but the number that identifies it is `bInterfaceNumber`, and the two are independent. The claim-failure message 20 lines above already interpolates `this._interfaceNumber`, so on a device where they differ the two errors disagree about the same interface.

This is exactly what Copilot flagged on the CH340 driver; fixed there in e8fd5ec, left out of that PR to keep the diff scoped.

**Fix:** interpolate `this._interfaceNumber`, matching CH340.

## FTDI: no endpoint validation at all

[`web/js/ftdi-webusb.js:139-146`](https://github.com/jasiek/webchirp/blob/master/web/js/ftdi-webusb.js#L139-L146)

Two differences from the other two drivers:

1. **No bulk-type filter.** The loop assigns on `direction` alone, with no `endpoint.type !== "bulk"` guard. This works today only because FTDI parts expose a bare bulk pair on interface 0 — PL2303 and CH340 both need the filter because they carry an interrupt IN endpoint for modem status, and any FTDI variant that did the same would silently bind the wrong endpoint.
2. **No guard after the loop.** If neither endpoint is found, `_inEndpoint`/`_outEndpoint` stay `0` and `open()` proceeds into control transfers, failing much later at `transferIn(0, ...)` with a message that says nothing about endpoints.

**Fix:** add the `type !== "bulk"` filter and the same post-loop guard the other two drivers have.

## Why it matters

None of this is reachable on the adapters verified so far, so this is robustness rather than a live bug. The value is diagnostic: these drivers run on Android where there is no dmesg and no serial console, and the in-app debug panel is the only signal a user can report. An error that names the wrong interface, or that surfaces as a generic transfer failure three steps after the real cause, costs a round trip on every bug report.

Worth doing as one pass so all three drivers handle endpoints identically — that consistency is what makes the pattern safe to copy for the next chip driver.

jasiek:
Closes #25.

The three WebUSB serial drivers selected and validated their bulk endpoints differently. This aligns the FTDI and PL2303 paths with CH340 so the pattern is safe to reuse for future chip drivers.

### FTDI (`web/js/ftdi-webusb.js`)

- Select only endpoints whose USB transfer type is `bulk`, so a later interrupt IN endpoint cannot replace the serial-data endpoint.
- Fail immediately when either bulk IN or bulk OUT is missing, before FTDI control initialization begins.
- Report the actual claimed interface number in the failure.

### PL2303 (`web/js/pl2303-webusb.js`)

- Report `this._interfaceNumber` when the bulk endpoint pair is missing instead of hard-coding interface 0.

### Tests

- Model endpoint transfer types in the FTDI fixture.
- Prove FTDI ignores an interrupt IN endpoint listed after bulk IN.
- Cover missing FTDI bulk IN and bulk OUT independently and verify validation precedes initialization.
- Cover the PL2303 diagnostic with a nonzero interface number.

## Impact

Existing verified adapters expose the expected bulk pair, so this is defensive hardening rather than a fix for a reproduced hardware failure. It prevents unusual endpoint layouts from silently selecting the wrong pipe and turns malformed layouts into an immediate, actionable diagnostic.

Rebased onto current `master` on 2026-08-17. No new dependencies.

## Validation

- `npm test` — syntax, 209 channel/UI/runtime tests, WebUSB, radio settings, and build tests passed
- `npm run test:webusb` — 54 passing, 0 failing

### DIFF
--- a/FINDINGS.md
+++ b/FINDINGS.md
@@ -72,6 +72,7 @@ CHIRP `b7ae1b6`.
 
 ## WebUSB serial adapters
 
+- **webusb-test-endpoints-must-name-their-type** (2026-08-17): WebUSB's `USBEndpoint` descriptors always identify their transfer `type`, but the original FTDI unit-test fake omitted it because that driver selected endpoints by direction alone. Once FTDI correctly filters for `type === "bulk"`, an untyped fake rejects every otherwise-valid `open()`. Driver fixtures must model descriptor type explicitly; the FTDI regression test also places an interrupt IN endpoint after bulk IN so the old direction-only selection demonstrably chooses the wrong pipe.
 - **ch340-version-not-bcddevice** (2026-07-28): the CH340/CH341 protocol branches on the chip version returned by vendor request `0x5F`, **not** on the USB `bcdDevice` word — an 0x1a86:0x7523 cable reporting `bcdDevice 0xc233` still reports a version of 0x30/0x31 over the wire. Version > 0x27 sets bit 7 of the prescaler value (otherwise the chip withholds data until a full 32-byte packet, which stalls short CHIRP replies); version >= 0x30 configures framing through the LCR register pair, and older parts are left at their 8N1 defaults.
 - **webusb-vendor-filters-are-too-broad-for-wch** (2026-07-28): FTDI and Prolific can be filtered by vendor id alone, but WCH's 0x1a86 also covers CH9102/CH343 parts that enumerate as CDC-ACM (and non-serial chips). The CH340 chooser filters on exact vendor/product pairs so those devices fall through to the CDC polyfill instead of getting the CH340 register map.
 
--- a/scripts/test-ftdi-webusb.mjs
+++ b/scripts/test-ftdi-webusb.mjs
@@ -6,7 +6,14 @@ import { FtdiSerialPort } from "../web/js/ftdi-webusb.js";
 // answers them, the way real hardware leaves a transfer outstanding until bytes
 // arrive. A driver that keeps a queue is only observable against a fake that
 // models the queue, so this cannot be a list of pre-baked results.
-function makeFakeDevice({ cancelOnClearHalt = false } = {}) {
+function makeFakeDevice({
+  cancelOnClearHalt = false,
+  endpoints = [
+    { type: "bulk", direction: "in", endpointNumber: 1, packetSize: 64 },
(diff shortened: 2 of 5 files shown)


jasiek:
Hey @lanrat - would you be able to find a minute or two to test if this still works on real hardware?

jasiek:
## Review — Approve ✅

The diff does exactly what #25 asked: after this lands, the endpoint-selection block in `ftdi-webusb.js`, `pl2303-webusb.js`, and `ch340-webusb.js` is structurally identical — claim by `bInterfaceNumber`, filter to `type === "bulk"`, assign by direction, and report the actual interface number on failure. No correctness defects found.

**Verified non-issues:**
- The new FTDI guard throws after `claimInterface`, but `BrowserSerialBridge.connect()` catches any `open()` failure and runs `_teardown()` (release + close, error-swallowing), and each retry constructs a fresh port instance — no leaked state.
- The bulk filter can't regress previously-working devices: all shipping FTDI serial parts expose exactly one bulk IN + one bulk OUT on the data interface, and endpoint number 0 can never be a bulk endpoint, so the truthiness guard is sound.
- `node --check` passes on all three changed/aligned files.

**Non-blocking notes:**
1. *(pre-existing, out of scope)* All three drivers still pick `configuration.interfaces[0]`. A composite device whose serial data interface isn't first would claim the wrong interface entirely — this PR correctly fixes only the reporting, per the scope of #25. For FTDI this interacts with the hard-coded `PORT_INDEX = 1`; multi-port FT2232/FT4232 parts would need real multi-interface support anyway.
2. *(nit)* The same block now exists verbatim in three drivers. A shared `selectBulkEndpoints(iface, chipName)` helper would prevent the drift this PR fixes from recurring.

jasiek:
Tested with FT232L on a Pixel 10.

PR #28, opened 2026-07-29
Load CHIRP images that do not identify their own driver

jasiek:
## Problem

CHIRP detects a radio from an image by walking `directory.DRV_TO_RADIO` and calling each driver's `match_model`, so **a driver that was never imported can never match**. webchirp imports drivers lazily, resolving which one to import from the image's metadata trailer.

Images written before CHIRP added that trailer name no driver at all. Dropping one into the app failed with `Unknown file format` unless the user happened to have already selected that exact radio. This affects a lot of real files — Baofeng UV-3R, most Yaesu (FT-60, FT-817, VX-*), Puxing, TYT TH-7800, Wouxun KG-UV8D, the Kenwood TK-* series.

Upstream doesn't hit this because `tests/test_drivers.py` opens with `directory.import_drivers()`, registering everything before detection runs.

## Change

Import every driver when — and only when — an image fails to resolve a driver up front, caching the result for the session. Modern images keep their existing fast path and import exactly one driver, so the cost is paid only by users opening legacy files, once per session.

Three cases now fall back to full detection:

1. **No metadata trailer** — the original motivating case.
2. **Metadata matching no catalog entry** — a guard against a driver the prebuilt catalog does not list. No image in `chirp/tests/images/` reaches it: every one of the 249 images carrying a trailer matches some catalog entry. An earlier revision of this description claimed this case fixed Quansheng UV-K5 egzumer; see **Known gap** below for why it does not.
3. **Metadata whose stored `rclass` resolves to a live-mode driver.** The Kenwood TS-480 image stores `rclass: "TS480Radio"`, and the catalog matcher prefers class-name matches, so it landed on `kenwood_live:TS480Radio` instead of the clone-mode `ts480:TS480_CRadio`. A live radio can never own a clone image, and `get_radio_by_image` only considers `FileBackedRadio` subclasses, so importing it could never help.

`import_all_driver_modules` reports which drivers failed to import rather than silently dropping them, as `list_registered_radios` does today, and surfaces them in the debug panel as `DRIVERS SKIP <module>`.

## Reconciled against master

Master has moved on since this branch opened; it is merged in here.

- **The pyserial shim (#38) fixed the two drivers this branch reported as unimportable.** `tg_uv2p` and `idrp` import pyserial at module scope and used to fail under Pyodide; with the shim the sweep imports all 191 driver modules and registers 551 radio classes with nothing skipped. The failure reporting stays — a submodule bump can reintroduce a failure, and importability can't be judged by grepping for `import serial`.
- **`scripts/build-catalog.mjs` now calls `import_all_driver_modules` instead of its own inline Python.** Master grew an inline loop doing exactly this sweep to warn about unimportable drivers; sharing the bridge function keeps the build's warnings and the runtime's skip lines from diverging, and keeps Python out of a JS file. Catalog output is byte-identical (551 radios, 191 modules).
- The one conflict was `package.json`'s `test:channels` list — both sides appended test files, so all of them are kept.

## Results

All 357 images in `chirp/tests/images/`, grouped by the path `handleLoadImage` actually takes for each, importing only what that path would import:

| Path | Images | Outcome |
|---|---|---|
| No metadata trailer → sweep | 108 | **load** — previously failed unless the user had already selected that exact radio |
| Metadata resolves to a live-mode driver → sweep | 1 | **loads** — `Kenwood_TS-480_CloneMode.img` as `ts480.TS480_CRadio` |
| Metadata resolves a driver → unchanged fast path | 248 | 247 load, 1 fails (below) |
| **Total** | **357** | **356 load** |

The previously reported holdout, `Quansheng_TG-UV2+.img`, now loads on the fast path: its driver imports under the pyserial shim and master added it to the radio catalog.

**Correcting an earlier revision of this description, which claimed 357/357.** That figure walked the corpus alphabetically inside a single runtime. `Alinco_DJ175.img` carries no trailer, so it triggers the sweep on the very first file, and every later image was then measured with all 191 drivers already imported — it measured "a session in which the sweep has already run", not each image's own path. The same artifact inflated the original 356/357.

## Known gap

`Quansheng_UV-K5_egzumer.img` does not load, in a fresh runtime with only that file:

```
resolved=uvk5.OSFWUVK5Radio  swept=false
chirp.errors.ImageMetadataInvalidModel: Unsupported model Quansheng UV-K5
```

Its metadata is `vendor=Quansheng, model=UV-K5, variant=egzumer`. `findCatalogRadioForImageMetadata` matches on vendor/model and ignores `variant`, so it resolves `uvk5.OSFWUVK5Radio` — a real, non-live catalog entry. That sets `resolvedDriver`, the sweep is skipped, and detection then fails. The correct driver is in a different module, `uvk5_egzumer.UVK5RadioEgzumer`; force the sweep and the image loads with 22 channels.

This is not a regression from the master merge — `OSFWUVK5Radio` was already in the catalog at the merge base, so the case never worked. The fix is to retry once through the sweep when detection fails after a fast-path resolve, which would also cover any future variant-in-another-module case without giving up the fast path. Not included here; this branch's scope is images that resolve no driver at all.

## Testing

New `scripts/test-metadataless-image-load.mjs`, wired into `npm run test:channels`, covering: metadata images needing no preload; a metadata-less image failing before the preload and loading after it (asserting channels are populated); the TS-480 live-driver clash; and failure reporting for unimportable drivers.

- `npm run test:channels` — 106 pass
- `npm run build:catalog` — 551 radios, no catalog diff
- `node --check web/app.js`, `node --check web/js/runtime-rpc.js`, `node --check scripts/build-catalog.mjs`

## Dependencies

None added.

https://claude.ai/code/session_017G138Fzv8BTL5nXcAYfiWr


jasiek:
## Review — Approve with nits ✅

The lazy full-driver import is correctly gated (only when neither the metadata trailer nor the catalog resolves a driver), cached per session with retry-on-failure, and safe against concurrent RPC calls since all runtime methods serialize through the call queue. Verified against upstream: `directory.get_radio_by_image` only considers already-registered `FileBackedRadio` subclasses, so both the live-radio skip and the preload are justified. Ran the branch in a disposable worktree: all 4 new tests pass, full `test:channels` passes 87/87, `node --check` / `py_compile` clean.

**Findings:**

1. **Minor — residual no-fallback path** (`web/js/runtime-rpc.js:293-295`): when metadata matches a non-live catalog entry by class name, the full import is skipped. If that driver's vendor/model/aliases then fail to match the trailer inside `get_radio_by_image` (class-name collision across modules, or a renamed model absent from `MODEL_COMPAT`), the load raises `ImageMetadataInvalidModel` with no retry. Catching detection failure and retrying after `ensureAllDriverModules()` would be strictly more robust. Not observed in the 357-image corpus.
2. **Minor — main-thread stall with no feedback** (`web/js/runtime-rpc.js:94-96`): the full import runs as one synchronous Python loop in a single `runPythonAsync` (~8.7 s in Node; longer in-browser since ~350 modules are fetched individually from jsDelivr). The only feedback is logged after completion — a session-status message before starting would help.
3. **Nit — misleading comment** (`web/js/runtime-rpc.js:283-286`): "let match_model pick the real driver" is wrong for the TS-480 case — the image has metadata, so `get_radio_by_image` matches vendor/model/variant against aliases and never calls `match_model` (which is only used for metadata-less files).
4. **Nit — stale FINDINGS.md**: `FINDINGS.md:31` asserts `idrp`'s `import serial` sits in dead paths and "imports fine"; this PR empirically found it fails under Pyodide (the import is at module scope, `chirp/chirp/drivers/idrp.py:16`). Worth amending here or in a follow-up.
5. **Note, not a defect**: metadata-less detection takes the first `match_model` hit in registration order, so ambiguous binary formats can theoretically misidentify — but this matches desktop CHIRP behavior after `import_drivers()` exactly; parity, not a regression.

jasiek:
Turns out loading all drivers isn't actually that bad and takes about one second.

PR #32, opened 2026-07-29
Preserve power levels and duplex when writing channels back

### LINKED ISSUE #31: Reading a radio and writing it back unchanged fails for 113 of 248 models
## Summary

Loading an image and writing it back — either uploading to the radio or exporting a `.img` — fails for **113 of the 248** upstream test images that resolve to a driver. Reading a radio and writing it back unchanged is not currently a safe operation.

This reproduces through the plain app path, with no test scaffolding:

```
Baofeng_UV-5R.img:  EXPORT FAILED -> InvalidDataError: Duplex is not +,-, or empty
BTECH_GMRS-V2.img:  EXPORT FAILED -> ValueError: 4.0W (36 dBm) is not in list
Kenwood_TH-D75.img: export OK (500637 bytes, 286 rows)
```

Breakdown of the 113:

| Class | Count |
|---|---|
| Power level round-trip | 67 |
| Duplex `split`/`off` rejected | 24 |
| Tone / DTCS validation | 5 |
| Settings re-apply rejected | 4 |
| Driver-specific tail (AttributeError ×8, NotImplementedError ×3, IndexError, KeyError) | 13 |

The first three (96 of 113) share one root cause. The rest need separate investigation.

## Root cause

`_apply_rows_to_radio_instance` (`web/python/runtime_bridge.py:591`) serializes rows back into CSV text and rebuilds each `Memory` through a legacy parser, rather than carrying the driver's own objects.

### Power — truncated dBm, then compared as floats

`_power_label_map_for_radio` (`web/python/runtime_bridge.py:270`):

```python
watts = chirp_common.dBm_to_watts(int(level))
formatted = str(chirp_common.AutoNamedPowerLevel(watts))
```

`int(level)` truncates the dBm. Measured on `anytone.AnyTone5888UVRadio`:

| Driver level | Real dBm | after `int()` | Our label | Reparsed dBm | Matches |
|---|---|---|---|---|---|
| High (50W) | 46.9897 | 46 | `39W` | 45.9106 | no |
| Mid1 (25W) | 43.9794 | 43 | `20W` | 43.0103 | no |
| Mid2 (10W) | 40.0 | 40 | `10W` | 40.0 | yes |
| Low (5W) | 36.9897 | 36 | `4.0W` | 36.0206 | no |

On write, `parse_power()` builds a fresh `AutoNamedPowerLevel`, the driver does `POWER_LEVELS.index(mem.power)`, and `PowerLevel.__eq__` compares float dBm — which no longer matches. The `ValueError: ... is not in list` is Python's `list.index()` message.

Only levels whose dBm is an exact integer survive (10W = 40, 1W = 30, 100W = 50), which is exactly why 67 images fail and the rest do not.

The truncation is the more dangerous half: a 50W level becomes the string `39W`. If the comparison ever succeeded we would be writing the wrong power to the radio. The failure is currently the only thing preventing silent corruption.

### Duplex and tone — a validator CHIRP itself routed around

Every write funnels through `Memory.really_from_csv()` (`runtime_bridge.py:624`), whose duplex handling is:

```python
if vals[3].strip() in ["+", "-", ""]:
    self.duplex = vals[3].strip()
else:
    raise errors.InvalidDataError("Duplex is not +,-, or empty")
```

The read path uses `mem.to_csv()`, which emits `split` and `off` verbatim — so **we emit values our own writer rejects**. The UV-5R test image has a `split` channel at location 1, which is enough to make export fail.

`really_from_csv` is legacy. CHIRP's own `generic_csv` driver does not use it; it has `_clean_duplex` handlers that set `split`/`off` directly. The function survives only in three older `chirp_common.py` call sites.

## No upstream test covers this

CHIRP has no byte-level image round-trip test — `grep` for `save_mmap|get_packed()` across `chirp/tests/*.py` returns nothing outside unit tests about metadata. The closest is `test_brute_force.py::test_get_set_all`, which round-trips at the `Memory` object level, and `assertEqualMem` (`chirp/tests/base.py:160`) explicitly skips power:

```python
elif k == "power":
    continue  # FIXME
```

So the exact field our largest failure class lives in is an annotated hole in upstream coverage, and duplex is never exercised through a CSV text round trip. There is no upstream oracle to inherit here.

## Proposed fix

Stop rebuilding `Memory` from text:

1. Resolve the power label against the radio's `valid_power_levels` and assign **the driver's own `PowerLevel` object**, so no dBm arithmetic or float comparison is involved.
2. Set duplex, tone and DTCS fields directly instead of going through `really_from_csv`, supporting the full set of values `to_csv()` can emit.
3. Fix the `int(level)` truncation so CSV export stops mislabelling 50W as 39W.

Add a load → export → compare test over `chirp/tests/images/`, which is coverage neither project currently has.

Found while measuring image-load coverage for #28.

jasiek:
Fixes #31.

## Problem

Reading a radio and writing it straight back failed for **113 of the 248** resolvable upstream test images. Reproducible through the plain app path — load a `.img`, click Export `.img`:

```
Baofeng_UV-5R.img:  EXPORT FAILED -> InvalidDataError: Duplex is not +,-, or empty
BTECH_GMRS-V2.img:  EXPORT FAILED -> ValueError: 4.0W (36 dBm) is not in list
```

Both came from `_apply_rows_to_radio_instance` rebuilding each `Memory` out of CSV text on the way back in.

## Power — truncated dBm, then compared as floats

`_power_label_map_for_radio` converted the driver's `PowerLevel` to a watt string via `int(level)`, which truncates the dBm:

| Driver level | Real dBm | after `int()` | Old label | Reparsed dBm | Matched |
|---|---|---|---|---|---|
| High (50W) | 46.9897 | 46 | `39W` | 45.9106 | no |
| Mid2 (10W) | 40.0 | 40 | `10W` | 40.0 | yes |
| Low (5W) | 36.9897 | 36 | `4.0W` | 36.0206 | no |

`PowerLevel.__eq__` compares dBm as a float, so the driver's `POWER_LEVELS.index()` lookup failed for every level whose dBm is not a whole number — exactly why 67 images failed and the rest passed.

Now the label resolves against the radio's own `valid_power_levels` and **that object** is assigned, so no arithmetic is involved. The truncation is also fixed for CSV export, where a 50W level had been written out as `39W`.

Unsupported power text now reports the radio's valid values instead of surfacing a bare `list.index()` error.

## Duplex, tone and DTCS — a validator CHIRP itself routed around

Every write went through `Memory.really_from_csv()`, which accepts only `+`, `-` and `""` for duplex — while the read path uses `to_csv()`, which emits `split` and `off`. **We emitted values our own writer rejected.** The UV-5R test image has a `split` channel at location 1, which was enough to break export.

`really_from_csv` is legacy; CHIRP's own `generic_csv` driver does not use it. Rows are now parsed directly, leaving the final say to the driver's `set_memory()` — the component that actually knows what it supports.

## Also fixed

A channel with no power level serializes as the string `"None"` (`to_csv` formats with `%s`). The old code could not parse that and fell back to a default, **silently writing the radio's first power level** onto such channels. It now stays unset.

## Also fixed: frequency precision

Row frequencies and offsets were converted with `to_MHz(float(text))`, copied from `really_from_csv`, which truncates — an 8.219000 MHz offset landed on 8218999 Hz, drifting 27 channels of the IC-M710 image by 1 Hz on every write. `parse_freq()` parses the same text without the float round trip.

## Results

Write failures across all 248 resolvable images: **113 → 25**.

Measuring fidelity rather than just success, across the 217 images that write: **212 now preserve every row**. 170 of those differ in bytes while every row is identical (padding and unused regions), so rows — not bytes — are the right oracle. Images whose channel data actually changes dropped from 8 to 5, tracked in #35; one of those five (`ftm7250d`) reproduces in pure CHIRP with no webchirp code involved.

The remaining 25 are unrelated to this change and stayed constant: 10 settings re-application, 3 the `DVMemory` column misalignment for D-STAR radios (`IC-2820H`, `ID-5100`, `TH-D74` — `DVMemory.to_csv()` has a different column layout than `Memory.to_csv()`, so `RxDtcsCode` receives the mode string), and a driver-specific tail of AttributeError/NotImplementedError/IndexError.

## Testing

New `scripts/test-channel-round-trip.mjs`, wired into `npm run test:channels`. **All 5 tests fail on master and pass with this change** (verified by stashing the fix):

```
not ok 1 - power levels survive a read/write cycle without dBm truncation
not ok 2 - channels with split and off duplex survive a read/write cycle
not ok 3 - power label resolution returns the driver's own PowerLevel objects
not ok 4 - a channel with no power level stays unset instead of getting a default
not ok 5 - unsupported power text fails with the radio's valid values, not an index error
```

They exercise real images through load → export → reload, asserting power labels and duplex/offset are unchanged across the write, plus unit-level coverage of the resolution mechanism itself.

Worth noting this is coverage CHIRP does not have: it has no byte-level image round-trip test, and `assertEqualMem` (`chirp/tests/base.py:160`) explicitly skips `power` with a `# FIXME`.

- `npm run test:channels` — 88 pass

## Dependencies

None added.

https://claude.ai/code/session_017G138Fzv8BTL5nXcAYfiWr


jasiek:
## Review — Approve with nits ✅

All four claims in the PR body check out against upstream sources: `PowerLevel.__int__` truncates dBm (`chirp_common.py:190`) so `dBm_to_watts(int(level))` renders 50 W as "39W" and the float-based `__eq__` then fails `POWER_LEVELS.index()`; `really_from_csv` accepts only `+`/`-`/`""` duplex while `to_csv` emits `split`/`off` and `"%s" % None` for unset power; and `to_MHz(float(...))` truncates where `parse_freq` does not. Column indexes in `_memory_from_row_values` match `Memory.CSV_FORMAT` exactly, both apply call sites go through the new parser, and the clone-workflow rules (cached image, prepared session) are untouched. The new tests are well-targeted (they assert `_resolved is _level` identity, the "39W" truncation, "None"/blank handling, and the 1 Hz drift) and correctly wired into `test:channels`.

**Findings:**

1. **Minor — tone/DTCS validation weakened** (`web/python/runtime_bridge.py`, `_memory_from_row_values`): `really_from_csv` rejected rtone/ctone not in `TONES` and dtcs/rx_dtcs not in `DTCS_CODES`; the new parser only checks float/int-ness. Many drivers encode these raw (e.g. `int(mem.ctone*10)`) without validating, so a malformed CSV fed via `npm run radio:write --format csv` can now pass `validate_rows_for_upload` and silently write a bogus tone into the image. UI dropdowns constrain the browser path, so exposure is mainly the agent-CLI/external-CSV path.
2. **Minor — negative offsets silently lose their sign**: upstream `parse_freq("-0.600000")` returns +600000 and `parse_freq("-5.600000")` returns −4400000 (the integer-MHz/kHz split mishandles negatives), whereas the old `to_MHz(float(...))` was sign-correct. No driver's `get_memory` emits negative offsets (they normalize to duplex `+`/`-`), so `to_csv`-derived rows are safe — but a hand-typed negative offset or imported CSV is now silently corrupted instead of round-tripping.
3. **Minor — unknown power labels now hard-fail upload**: `_resolve_power_level` raises where `_normalize_power_value` previously fell back to a default. "Fail clearly" is the right call, but importing a CSV exported from a *different* model (labels like "L1"/"Mid") now blocks upload entirely — worth a line in the PR description.
4. **Nit — CSV-text export still assigns default power**: `normalize_rows` still routes Power through `_normalize_power_value`/`DEFAULT_EXPORT_POWER`, so an unset-power channel exports to CSV as "50W". The "no silent default" fix covers only radio/image writes. Pre-existing, not a regression.
5. **Nit — error-column inference**: "rTone is not a valid number" maps to column "Tone" (keyword "tone" matches before "rtonefreq" in `_infer_csv_error_column`). Same misattribution existed with the old upstream messages.

PR #36, opened 2026-07-30
Make CSV export byte-identical to CHIRP's own CSV export

jasiek:
Answers the two questions from #32's review, and fixes what they turned up.

**Does CHIRP export numeric power values?** Yes — always watts, never a driver label. Export goes through `generic_csv.CSVRadio.set_memory()`, which rewraps every level as `AutoNamedPowerLevel(dBm_to_watts(float(level)))`, so `High` on an AnyTone 5888UV is written `50W` and `Low` on a UV-5R is `1.0W`.

**Were we in line with CHIRP?** No — **318 of the 356 loadable upstream test images exported differently**, and some of the differences lost data. This PR takes that to 5, all of which differ only in the Power column.

## What CHIRP actually does

`ChirpMemEdit._export_to_file()` (`chirp/wxui/memedit.py:2758`) pushes each memory through `import_logic.import_mem(csv_radio, rf, mem, mem_cls=chirp_common.Memory)` into a `CSVRadio` and saves that. We were writing rows straight to CSV text, so we missed everything that path does.

## Data loss fixed

**A phantom channel 0 on every import and export** — the single biggest cause, present in any export whose channels do not start at location 0. `CSVRadio(None)` seeds a *non-empty* channel 0 at 146.010000/50W, and `CSVRadio._load()` — the path `load_from()` takes — never calls `_blank()`; only the filename path `load()` does. Upstream never trips over it because `memedit.py:2774` calls `erase_memory(0)` explicitly.

**Every D-STAR channel dropped.** `DVMemory.to_csv()` (`chirp/chirp_common.py:614`) still emits the pre-RxDtcsCode/CrossMode/Power layout — 18 fields where `Memory.CSV_FORMAT` has 21 — so zipping it against the headers put the mode string in `RxDtcsCode`. The grid showed `RxDtcsCode=DV`, and export then failed those rows outright (`[rx_dtcs] invalid literal for int(): 'DV'`). `Icom_IC-2820H` exported 12 channels in CHIRP and 0 here. CHIRP dodges it by forcing `mem_cls=chirp_common.Memory`; we now do the same. Call signs are dropped either way — CHIRP's CSV export drops them too.

**Channels with no frequency dropped.** `CSVRadio._load()` skips any row whose frequency parses to 0, but drivers do report non-empty memories with no frequency (`Icom_IC-W32A`, `Jetstream_JT220M`) and CHIRP exports them, because its export never round-trips through its own parser. Rows now go through `_parse_csv_data_line()` per row, which keeps CHIRP's `ATTR_MAP` converters and its 50W default without the `_load()` filter.

## Correctness fixed

**Power was truncated.** `dBm_to_watts(int(level))` — `PowerLevel.__int__` truncates the dBm, so an AnyTone `High` (46.99 dBm) was written `39W` and `Mid1` `20W` instead of `50W`/`25W`. Now `float()`. This overlaps one line with #32, which fixes the same truncation for the radio-write path.

**Tone and DTCS columns.** `_import_tone`/`_import_dtcs` fill in columns the CSV format carries but a radio may not track separately — rtone into ctone, dtcs into rx_dtcs. FT-60 DTCS channels export with `RxDtcsCode=263` in CHIRP and `023` without that step.

**Features came from a blank driver instance.** `Rt98Radio` advertises only the PMR level (`Low` = 0.5W) until an image is parsed, then the full `Low`/`Mid`/`High` set where `Low` = 5W — so a label resolved against blank features picked the wrong wattage. `_driver_features()` now prefers the cached image, which is where CHIRP's features come from. Same root cause as the existing **settings-need-backing-state** finding.

**Incompatible memories.** CHIRP logs and writes them unconverted rather than dropping them; we now match, and the reason lands in the debug panel.

## Results

Every `.img` in `chirp/tests/images` (357), exported both ways in the same interpreter:

| | master | this PR |
|---|---|---|
| Byte-identical to CHIRP | 38 | **351** |
| Differ in channel data | 318 | **0** |
| Differ only in the Power column | — | 5 |
| Image fails to load at all | 1 | 1 |

The 5 residual cases are drivers that use one label for two wattages — `vx6`/`vx7.POWER_LEVELS_220` reuse `Hi`/`L3`/`L2`/`L1`, `ft90` splits its labels across VHF and UHF, and `radtel_t18.RB618Radio`/`RT647Radio` assign a `Low` they never advertise. A row carries power as text, so the label alone cannot say which list it came from; CHIRP is exact only because it passes `PowerLevel` objects around. Radio *writes* are unaffected (`vx6.set_memory` re-resolves power from the frequency).

Worth recording why the fix isn't "show watts in the grid instead": across all 551 driver classes, **zero** advertise two levels sharing a label, but **8** collapse onto the same watt string once `dBm_to_watts()` rounds to 1 dp — `retevis_rt21.RT619Radio`, `radtel_t18.RB619Radio`, `bf_t8.RetevisRB627B` and others put High *and* Low on `0.5W`. Watts in the grid would make those indistinguishable and unresolvable on the way back, and would diverge from CHIRP's UI, which shows the label.

`Quansheng_TG-UV2+.img` still fails to load (`ImageMetadataInvalidModel: Unsupported model Quansheng TG-UV2+`); that is driver detection, unrelated to CSV, and unchanged.

## Testing

New `scripts/test-csv-export-parity.mjs`, wired into `npm run test:channels`. It diffs our export against a transcription of `memedit._export_to_file()` running in the same interpreter, so the oracle is upstream's code rather than a restatement of ours. 20 curated images by default (~27 s), one per divergence class plus a vendor spread; `WEBCHIRP_CSV_PARITY_IMAGES=all` sweeps all 357 (~3 min). The 5 Power-column cases are listed explicitly, and the test fails both if that set grows *and* if a listed image starts matching. Two unit tests cover the import direction — no phantom channel 0, and a DV row keeps its column layout.

- `npm run test:channels` — 87 pass
- `npm run test:settings` — 551 pass
- `WEBCHIRP_CSV_PARITY_IMAGES=all` — 4 pass, 351 images identical (318 diverge on master with the same test)
- `python -m py_compile web/python/runtime_bridge.py`, `node --check` on the touched JS

## Dependencies

None added. `chirp/import_logic.py` joins the eagerly-seeded core CHIRP files (`web/js/python-sources.mjs`) because `runtime_bridge.py` imports it above the lazy import hook; it is 343 lines and its only imports are `chirp_common` and `errors`, both already seeded.


jasiek (commented):
## Review

Verified rather than just read: the reference export in the test is a faithful transcription of `ChirpMemEdit._export_to_file()` (memedit.py:2758–2788) including `erase_memory(0)`, `mem_cls=Memory`, the specials filter, and writing rejected memories unconverted. The `CSVRadio` claims hold (`_load()` skips `_blank()` and drops zero-frequency rows; `set_memory()` rewraps power via `dBm_to_watts(float(...))` and `_grow()`s, so the `max_memory` difference vs upstream is immaterial to output). `csv`/`io` imports are removed with no remaining uses, and `import_logic.py` imports only `logging`, `chirp_common`, `errors`, so the eager-seeding claim is accurate.

Ran the branch in a clean worktree: `npm run test:channels` → 87 pass; the parity test alone reports exactly the 5 known Power-column divergences and passes; `node --check` and `py_compile` clean.

**Strengths**: the parity oracle is upstream's own code, not a restatement of ours; the known-divergence list is enforced in both directions; the curated image list is guarded against submodule renames; the `compared >= 10` floor prevents a silently-skipping run from reading as green. `_blank_csv_radio()` gives future `CSVRadio(None)` call sites an obvious right answer.

**Verdict**: LGTM — the fix removes real data loss (phantom channel 0, dropped D-STAR and zero-frequency channels), residual divergences are honestly bounded and pinned by tests, no dependencies added. Inline comments are all minor; only the radio-read robustness one is worth a follow-up. (Posting as a comment review since GitHub won't let the PR author approve.)

On `CSVRadio._parse_csv_data_line()`: coupling to a private upstream API, but the rationale is sound and the parity test will catch any submodule bump that changes its behavior. Acceptable.

web/python/runtime_bridge.py
```diff
@@ -279,33 +347,98 @@ def _coerce_csv_vals_for_chirp(vals):
...
+
+    memories = []
+    for index, row in enumerate(rows or []):
+        vals = [str(row.get(header, "") or "") for header in CSV_HEADERS]
+        vals = _coerce_csv_vals_for_chirp(vals)
+        vals[power_idx] = _csv_export_power_text(vals[power_idx], power_map)
+        try:
+            mem = parser._parse_csv_data_line(list(CSV_HEADERS), vals)
+        except Exception as exc:
+            raise RuntimeUnsupportedError(f"Channel row {index + 1}: {exc}") from exc
```
jasiek, line 420: **Minor — a malformed row now aborts a completed radio read.** The old path collected row errors in `radio.errors` and skipped the row; this raises on any unparseable row. For grid-driven export that's an improvement (no silent channel loss). But `normalize_rows()` also runs at the end of the read-from-radio RPC (`csv_text = normalize_rows(rows, ...)` right after `sync_in()`), so one row the CSV converters choke on would discard an otherwise successful clone. Risk is low since those rows come from `mem.to_csv()`, but consider making `csvText` best-effort in the read path so a clone is never thrown away over one row.

web/python/runtime_bridge.py
```diff
@@ -222,26 +266,45 @@ def parse_csv(csv_text: str):
...
+        return None
     try:
         radio_cls = _import_radio_class(module_name, class_name)
+    except Exception:
+        return None
+
+    image = LAST_IMAGE_BY_DRIVER.get(_driver_cache_key(module_name, class_name))
+    factories = [lambda: radio_cls(None), lambda: radio_cls("")]
+    if image:
+        factories.insert(0, lambda: radio_cls(memmap.MemoryMapBytes(bytes(image))))
```
jasiek, line 288: **Minor — re-parses the cached image on every call.** `validate_rows_for_upload()` and `_apply_rows_to_radio_instance()` now construct a driver from `LAST_IMAGE_BY_DRIVER` (full `process_mmap`) per call, where they previously used a blank instance. That's the correctness fix working as intended, but if validation is invoked per-edit from the UI it's a repeated image parse. Also `_apply_rows_to_radio_instance` already holds a parsed `radio` instance and could take `radio.get_features()` directly instead of re-deriving from module/class names. Fine to defer unless validation latency shows up.

web/python/runtime_bridge.py
```diff
@@ -279,33 +347,98 @@ def _coerce_csv_vals_for_chirp(vals):
...
-    writer.writerow(CSV_HEADERS)
-    for row in rows:
-        cooked = [row.get(header, "") for header in CSV_HEADERS]
-        cooked = _coerce_csv_vals_for_chirp(cooked)
-        power_idx = CSV_HEADERS.index("Power")
-        cooked[power_idx] = _normalize_power_value(
-            cooked[power_idx], power_map, default_power
-        )
-        writer.writerow(cooked)
+def _csv_export_power_text(value, power_map):
```
jasiek, line 350: **Nit — near-duplicate of `_normalize_power_value()`**, differing only in the blank/fallback handling (export always falls to `50W`; the write path prefers the driver's first level). The split is deliberate, but a shared helper with a fallback parameter would keep the two from drifting apart silently.

web/python/runtime_bridge.py
```diff
@@ -279,33 +347,98 @@ def _coerce_csv_vals_for_chirp(vals):
...
+    frequency parses to 0, but drivers do report non-empty memories with no
+    frequency (uninitialised locations in ``Icom_IC-W32A``, ``Jetstream_JT220M``)
+    and CHIRP exports those, because its export never round-trips through its own
+    parser. Everything else about the conversion stays CHIRP's, including the
+    ATTR_MAP converters and the 50W default for a channel with no power.
+    """
+    parser = _blank_csv_radio(0)
+    # _clean_tmode() mirrors one tone onto the other when the file carries only
+    # one of the two columns; rows always carry both.
+    parser.file_has_rTone = True
```
jasiek, line 408: **Nit — currently a no-op**: `_clean_tmode()` behaves identically when both flags are `None` (both falsy) as when both are `True`. Harmless, and it does document intent against upstream changes to `_clean_tmode`; just noting it isn't load-bearing today.

scripts/test-csv-export-parity.mjs
```diff
@@ -0,0 +1,308 @@
...
+  if (expectedLines.length !== actualLines.length) {
+    return false;
+  }
+  let sawPowerDifference = false;
+  for (const [index, line] of expectedLines.entries()) {
+    const other = actualLines[index];
+    if (line === other) {
+      continue;
+    }
+    const mine = line.split(",");
```
jasiek, line 132: **Nit — naive `","` split**: a quoted Comment containing a comma would misalign fields. The failure mode is conservative (field-count mismatch → treated as a real divergence), and it only applies to the 5 known-divergence images, none of which carry commas in comments. Not worth fixing unless it ever fires.

web/python/runtime_bridge.py
jasiek: We'll see if it becomes a performance issue.

PR #38, opened 2026-07-30
Ship a pyserial shim so serial-importing drivers register (fixes #29)

### LINKED ISSUE #29: Drivers importing pyserial at module scope are silently unavailable (Quansheng TG-UV2+, Icom idrp)
## Summary

Two CHIRP drivers import `serial` (pyserial) at module scope. Pyodide has no `serial` module, so importing them raises `ModuleNotFoundError` and they never register in `directory.DRV_TO_RADIO`:

```
idrp:    ModuleNotFoundError: No module named 'serial'
tg_uv2p: ModuleNotFoundError: No module named 'serial'
```

Consequences:

- Neither radio appears in `web/radio-catalog.json`, so they cannot be selected in the UI at all (confirmed: zero entries matching `TG-UV2`).
- `chirp/tests/images/Quansheng_TG-UV2+.img` cannot be loaded. It is the only image out of 357 in the upstream corpus that still fails after #28 — everything else loads.

The image fails with a misleading message, because detection falls through to the "no registered class matched" branch:

```
RuntimeUnsupportedError: Unable to detect radio from image: Unsupported model Quansheng TG-UV2+
```

## Why it was invisible

`list_registered_radios` in `web/python/runtime_bridge.py` catches per-module import errors and `continue`s, so a driver that cannot import is indistinguishable from one that does not exist. #28 adds `import_all_driver_modules`, which reports failures to the debug panel, so these two are now at least visible there — but the catalog build still drops them silently.

## Possible directions

1. Ship a minimal `serial` shim module in the Pyodide runtime that satisfies module-scope imports. Both drivers are clone-mode over a pipe, so they may only need the module to exist for import to succeed, not to function.
2. Report unimportable drivers during catalog build (`scripts/build-catalog.mjs`) instead of dropping them, so the gap is recorded rather than inferred.
3. Give the image-detection failure a clearer message when metadata names a vendor/model whose driver is known to be unimportable, rather than "Unsupported model".

## Reproduction

```js
// with the harness from scripts/test-radio-harness.mjs
await harness.runPythonJson(`
_out = {}
try:
    importlib.import_module("chirp.drivers.tg_uv2p")
    _out["import"] = "ok"
except Exception as e:
    _out["import"] = "%s: %s" % (type(e).__name__, e)
json.dumps(_out)`)
// => { import: "ModuleNotFoundError: No module named 'serial'" }
```

jasiek:
## Summary

Fixes #29. Two CHIRP drivers (`tg_uv2p`, `idrp`) import pyserial at module scope; Pyodide has no `serial` module, so both raised `ModuleNotFoundError`, never registered in `directory.DRV_TO_RADIO`, were absent from the radio catalog, and `chirp/tests/images/Quansheng_TG-UV2+.img` failed detection with a misleading "Unsupported model" error.

- **`web/python/runtime_bridge.py`**: installs a shim `serial` module before any driver imports (issue direction 1). It carries pyserial's public constants (`STOPBITS_*`, `PARITY_*`, byte sizes) plus `SerialException`/`SerialTimeoutException`, and a `Serial` class that refuses construction with a clear error — radio I/O always goes through `WebSerialPipe`, so the constants are all the drivers actually need (`tg_uv2p` uses `serial.STOPBITS_TWO` at clone time; `idrp` only touches `serial.Serial` in its `do_test()` CLI helper).
- **`scripts/build-catalog.mjs`**: warns per driver module that fails to import instead of dropping it silently, so a catalog gap is recorded rather than inferred (issue direction 2). Currently zero of 191 modules fail.
- **`web/radio-catalog.json`**: regenerated — Quansheng TG-UV2+ now appears (551 radios, up from 550).
- **`scripts/test-pyserial-shim.mjs`** (wired into `test:channels`): pins that both drivers import, TG-UV2+ registers and its upstream test image detects the correct driver with channel rows, and the shim `Serial` fails loudly.

Note: `idrp.IDRPx000V` has no `@directory.register` decorator upstream, so it does not appear in the catalog even in desktop CHIRP — fixing its import is all that can be done for it here (recorded in FINDINGS.md).

## Validation

- `node --check web/app.js`, `node --check web/js/runtime-rpc.js`, `python3 -m py_compile web/python/runtime_bridge.py`
- `npm run test:channels` — 102/102 pass (including the 4 new shim tests)
- `npm run build:catalog` — 551 radios, zero unimportable modules

## New dependencies

None.


jasiek:
`Quansheng_TG-UV2+.img` confirmed to be loadable via UI.

jasiek (commented):
Reviewed the shim, the catalog-build warning logic, the new tests, and the regenerated catalog. Overall this looks solid and correctly fixes issue #29 — no blocking findings.

Verified locally:
- Shim fidelity: constants match pyserial's `serialutil` exactly (FIVEBITS..EIGHTBITS, PARITY_* chars, STOPBITS_ONE/ONE_POINT_FIVE/TWO), and the exception hierarchy matches (`SerialException(OSError)` — pyserial's base is `IOError`, an alias of `OSError` — with `SerialTimeoutException(SerialException)`).
- Only `tg_uv2p` and `idrp` import pyserial at module scope in `chirp/drivers/` (the others — `ic9x`, `hf90`, `tmv71_ll` — import inside `_test`/`__main__` blocks), so the shim's scope claim is accurate. `tg_uv2p` only needs `STOPBITS_TWO`; `idrp` constructs `Serial` only in its `__main__` test path, so the raise-on-construction stand-in cannot misfire in the web runtime.
- Ordering is right: the shim is installed before `from chirp import ...` and before any driver import, and the `import serial` guard means a real pyserial (e.g. micropip-installed) is never masked.
- Ran `node --test --experimental-wasm-stack-switching scripts/test-pyserial-shim.mjs`: 4/4 pass in real Pyodide, including the upstream `Quansheng_TG-UV2+.img` load. Syntax checks (`node --check`, `py_compile`) pass.
- Regenerated catalog checks out: 551 radios, `count` field consistent, `tg_uv2p:QuanshengTGUV2P` present, and `idrp` correctly absent (confirmed upstream `IDRPx000V` has no `@directory.register`, matching the FINDINGS entry).
- Commits are small and scoped per repo conventions; Python stays in `web/python`.

Three non-blocking notes are attached inline: a real follow-up about `stopbits` never reaching the Web Serial port config, a possible misleading error message in the catalog failure scan for partially-registering modules, and a displaced eslint-disable comment.

🤖 Review by Claude Code sub-agent

web/python/runtime_bridge.py
```diff
@@ -31,6 +32,62 @@ def _install_gettext_builtins() -> None:
 
 _install_gettext_builtins()
 
+
+def _install_pyserial_shim() -> None:
+    """Provide a stand-in ``serial`` module so pyserial-importing drivers load.
+
+    Pyodide has no pyserial, so drivers that ``import serial`` at module scope
+    (tg_uv2p, idrp) would fail to import and never register with CHIRP's
+    directory. They only need the module's constants — radio I/O always goes
```
jasiek, line 41: Non-blocking follow-up: "they only need the module's constants" is true for import-time, but at clone time `tg_uv2p.do_ident` does `radio.pipe.stopbits = serial.STOPBITS_TWO`. `WebSerialPipe` silently absorbs that as a plain instance attribute, and `web/js/serial.js` hardcodes `stopBits: 1` in `port.open(...)`, so a live TG-UV2+ clone would run at 1 stop bit while the protocol expects 2 — ident will likely fail on real hardware. That's out of scope here (the driver was previously absent entirely, and image loading works — the tests prove it), but it's worth a FINDINGS.md note or follow-up issue so the first live-clone bug report isn't a surprise: plumbing `stopbits` through `serial_open`/`WebSerialBridge` would be needed to finish the job.

scripts/build-catalog.mjs
```diff
@@ -36,6 +36,24 @@ async function main() {
...
   );
 
+  // list_registered_radios drops unimportable modules silently; surface them
+  // here so a driver missing from the catalog is recorded, not inferred.
+  const importFailures = await harness.runPythonJson(
+    `
+_failures = {}
+for _name in _modules:
+    try:
+        importlib.import_module("chirp.drivers." + str(_name))
```
jasiek, line 46: Edge case, non-blocking: `list_registered_radios` already attempted (and swallowed) the import of every module, so this retry re-executes the module body for anything that failed. If a module registers some radios and *then* raises, the re-import will die on `Duplicate radio driver id` (the re-execution behavior recorded in FINDINGS under rpc-fifo-queue) rather than the original error — the module is still correctly flagged, but the reported reason could be misleading. Not worth restructuring for now; a one-line comment noting the caveat would do.

scripts/build-catalog.mjs
```diff
@@ -57,8 +75,9 @@ async function main() {
 
   await writeFile(OUTPUT_PATH, `${JSON.stringify(catalog, null, 0)}\n`, "utf8");
   // eslint-disable-next-line no-console
+  const failureCount = Object.keys(importFailures).length;
```
jasiek, line 78: Nit: the inserted `const failureCount` line now sits between the `// eslint-disable-next-line no-console` comment and the `console.log` it was annotating, so the suppression points at the wrong line. Cosmetic only (the repo has no eslint config), but worth swapping the two lines if you touch this file again.

PR #41, opened 2026-07-30
Show progress while loading CHIRP drivers, and unify the app font

jasiek:
## Problem

Loading an image that identifies no driver triggers the all-drivers sweep from #28: 191 driver modules, each a separate jsDelivr fetch in the browser, roughly **20 seconds**. It ran with no indication whatsoever — the app simply stopped responding, which is indistinguishable from a hang.

There was nowhere for it to report to. `log.setStatus()` only appends a `STATUS` line to the debug textarea, and the one progress widget that exists — `#clone-progress` — lives inside a collapsible `<details>` in the sidebar, so it is invisible whenever that section is closed.

## Change

**An app-wide progress strip**, rendered as the first row inside the Debug Output panel, where the matching narration already lands. It carries no border or background of its own so the panel still reads as one surface, and it is `hidden` whenever nothing is running.

`createProgress({ dom }).begin(label, max)` returns an `{update, end}` handle keyed to the operation that opened it, so a slow operation finishing late cannot repaint or tear down a strip a newer one owns. The bar goes indeterminate rather than inventing a percentage when no total is known, and clamps a count that overshoots its total.

**Progress from the driver sweep.** `import_all_driver_modules()` takes an optional `progress_cb` and calls it after each module. `runtime-rpc.js` drives the strip from it and narrates every 25th module to the debug panel:

```
IMAGE metadata absent; importing all drivers for detection
DRIVERS image identifies no driver; importing all 191 driver modules so match_model can run
DRIVERS 25/191 imported (latest ft4)
DRIVERS 50/191 imported (latest hf90)
...
DRIVERS 191/191 imported (latest yaesu_clone)
DRIVERS imported 191/191 modules, 551 radio classes registered
```

One debug line per driver would bury every other diagnostic; none at all leaves a stalled sweep looking identical to a working one. `DRIVER_LOG_INTERVAL` in `runtime-rpc.js` is the knob.

The strip is torn down in a `finally`, so a failed sweep does not leave a frozen bar on screen for the rest of the session.

## Why a callback and not per-module RPCs

A synchronous Python `for` loop inside a single `runPythonAsync` can animate browser UI, which looks impossible. Every CHIRP import suspends the interpreter on a CDN fetch through `ChirpCdnFinder`'s JSPI `run_sync`, so the event loop runs between iterations and DOM writes paint — the same mechanism that lets CHIRP's synchronous clone loops drive `#clone-progress` through `serial_progress`. Restructuring the sweep into per-module RPCs would have paid a full call-queue round trip per module for no benefit. Recorded in `FINDINGS.md` as **synchronous-python-can-drive-progress**.

## Also: one font for the whole app

Noticed while checking how the new strip's type sat against the rest of the UI — the page had **three** fonts, not two. `body` set Segoe UI and `#debug-output` set a monospace stack, but form controls do not inherit `font-family`, so every button, dropdown, `option` and text input fell back to the user agent default: 159 elements with an empty channel grid, far more with a codeplug loaded, since the grid is built from `select` and `input`. That default is Arial on macOS and *isn't* Arial everywhere, so the mismatch varied by platform.

The app is now Arial by declaration rather than by accident:

```css
button, input, select, textarea { font-family: inherit; }
body { font-family: Arial, Helvetica, sans-serif; }
```

Only the family is inherited, not the `font` shorthand — the shorthand would also pull in the body `font-size` and grow every control from the UA's 13.33px to 16px. Measured against master in the browser to confirm nothing else moved: Import CSV 95×44, Download Radio 85×46, radio-make 44 tall, all at Arial 13.33px, identical on both sides. `#debug-output` keeps its monospace stack through id specificity. Computed families on the live page are now exactly two.

`about.html` shares the stylesheet, so it picks up Arial as well.

## Accessibility

`aria-live` sits on the label, not the container. The label changes once per operation while the count ticks once per module, so a live region around both would announce ~191 times during a single sweep. The `<progress>` element carries the numbers for anyone who asks for them.

## Testing

New `scripts/test-progress.mjs`, wired into `npm run test:channels` — 7 tests covering hidden-until-begun, count and percentage, overshoot clamping, the indeterminate case, stale-handle isolation, per-module callback ordering (1-indexed, failures included), and that a throwing callback cannot abort the sweep while the argument stays optional.

- `npm run test:channels` — 113 pass
- `node --check web/app.js`, `web/js/runtime-rpc.js`, `web/js/ui.js`, `web/js/ui/progress.js`, `web/js/ui/dom.js`
- **Verified in a real browser** against `chirp/tests/images/Baofeng_UV-3R.img`: strip hidden at rest, ticks `0 / 191` → `52 / 191` → `175 / 191` over the sweep, tears down on completion, image loads with 3 channels. A second load in the same session reuses the cached sweep and shows no strip, as intended.

## Base branch

Targets `claude/preload-drivers-metadataless-images` (#28), not `master`: this reports on `ensureAllDriverModules()` / `import_all_driver_modules()`, which only exist there. Retarget once #28 lands.

## Dependencies

None added.


jasiek:
Considered loading modules incrementally, but loading them all is so quick we don't need to worry about that for now.

PR #47, opened 2026-07-31
Add an RSGB ETCC repeater query driven by GPS position

jasiek:
Adds a **🇬🇧 Query RSGB ETCC API** action that turns a GPS fix into a list of nearby UK repeaters, ready to program.

The position becomes a Maidenhead locator, the distance decides which locator squares to fan out over, and distance, band, mode and status are applied client-side before the survivors land in the channel grid nearest-first. Country is fixed to the United Kingdom — the directory is UK-only, and a picker with one entry is a control that cannot do anything.

Unlike the przemienniki.net and RepeaterBook imports, this needs **no CORS proxy**, so it works on deployments where those two are disabled.

## Why the code looks the way it does

Four behaviours of the upstream API shaped it. All are documented in `FINDINGS.md` and pinned by tests:

- **`/locator` only prefix-matches at four characters.** At six it is exact string equality, so `/locator/IO83DC` misses the record filed at `IO83DC20`, and a 6-character query cannot reach the ~26% of records stored at coarser precision. The fan-out therefore never queries the 6-character square the user is standing in.
- **A miss is HTTP 200 with `{"data":null}`**, not an HTTP error. Callers must check the payload, not the status.
- **Beacons report `rx` as 0.** All 36 of them are transmit-only, so "is this a repeater" is `tx > 0 && rx > 0 && tx !== rx`. A bare `tx !== rx` would admit every beacon as a duplex repeater with a ~145 MHz offset.
- **There are no coordinates in the payload.** Position comes from the locator alone, at mixed precision (4, 6 and 8 characters all occur, plus one invalid 5-character value).

## Three bugs caught by running it in a browser

Worth calling out, since they only appeared against live data:

1. **Empty coordinate fields read as 0°N 0°E** — `Number("")` is `0`, not `NaN`, so an untouched modal claimed to be in `JJ00AA`.
2. **Rows outside the requested distance.** Filtering on distance to the nearest corner of a record's locator box returned rows reading 33.6 km from a 30 km search, and ranked any station inside the searched square at 0.0 km — floating an anywhere-in-130 km record above a repeater measured at 13.6 km. Now filtered and ranked on the box centre, so the number shown always agrees with the filter.
3. **Half-built junk rows.** GB3EN is a 1312 MHz ATV repeater; on a 2m/70cm handheld the grid rejected the frequency but accepted the −63 MHz offset. Those rows are now dropped and the count reported in the status line.

## Scope decisions a reviewer might question

- **Only repeaters.** Simplex gateways, hotspots, packet nodes and beacons are excluded, and the mode list carries only the eight modes a repeater can be in. ATV repeaters still reach the grid (they are genuinely duplex) but have no checkbox of their own.
- **Bands are the seven that hold a repeater**, ordered by repeater count so 70cm and 2m lead with 92% of them. Two different reasons for the exclusions: HF (40/30/20/15m) has no repeater allocation at all and is permanently empty, whereas 4m/13cm/6cm/24GHz/SHF are repeater-capable allocations that simply have none coordinated today. **The second half is a snapshot and will go stale** — a future 4m repeater would still reach the grid (an empty band selection means "any band") but could not be selected on its own. `npm run test:api` alarms on exactly this.
- **Filters reset to their defaults on every open** rather than remembering the last query, so the modal always opens in the state it documents.

## Testing

- **`scripts/test-rsgb.mjs`** — 23 unit tests over locator maths, fan-out planning, dedup, filtering, ranking and row building.
- **`scripts/test-ui-rsgb-query.mjs`** — 14 tests driving `createRsgbQuery` against a fake DOM (a real `classList` and a `querySelectorAll` that understands `:checked`, since the module reads state back out of the DOM).
- **`scripts/test-rsgb-live.mjs`** — 12 contract tests against the live API, run with `npm run test:api`. **Deliberately not in `npm test`**: it needs the network and a third party's uptime. The rest of the suite runs on fixtures transcribed from this API, so it stays green if the upstream changes shape — these are the only thing that catches that. Mutation-checked (dropping `2M` from `RSGB_BANDS` fires the staleness guard).

`npm test` is 743 tests, all passing. `npm run test:api` is 12/12 against the live API.

Also verified by hand in the browser against the live API: a 30 km query from central London (`IO91WM`) fans out over `IO91` + `JO01`, and the default filters return 8 repeaters — GB3XP 145.6875 −600 kHz at 13.6 km through GB3ML 430.9125 +7.6 MHz at 29.7 km, all matching the UK bandplan. Real geolocation is blocked in the automation browser, so the denial path was verified for real and the success path with a stubbed fix; **the GPS button still wants one human click to confirm end to end.**

## Dependencies

**None added.** No `package.json` dependency changes — the only edits there register the two new test files and add the `test:api` script.


web/js/rsgb.js
```diff
@@ -0,0 +1,494 @@
...
+  ];
+  for (const [flag, choices] of digital) {
+    if (flags.has(flag)) {
+      const match = findEnumOption("Mode", choices, true);
+      if (match) {
+        return match;
+      }
+    }
+  }
+  return findEnumOption("Mode", analogue, true);
```
jasiek: **P1 — Preserve the selected digital mode instead of falling back to analogue.** The mode checkboxes only filter inclusion; `buildRsgbRows()` does not receive that selection. As a result, a mixed A/DMR repeater selected via a DMR-only query is still written as NFM, and a DMR-only repeater on an analogue-only radio also falls through here to NFM. Neither produces a usable DMR channel. Please pass the requested modes into row construction and choose one the radio supports, or skip and report incompatible digital rows rather than silently converting them to analogue.

web/js/ui/rsgb-query.js
```diff
@@ -0,0 +1,246 @@
...
+      return;
+    }
+    if (plan.truncated) {
+      // A clipped plan queries a subset of the area, so say so rather than
+      // letting a short result read as "that is everything nearby".
+      log.logDebug(`RSGB PLAN truncated to ${plan.squares.length} of ${plan.considered} squares`);
+      log.setStatus(`Distance spans ${plan.considered} squares; querying the ${plan.squares.length} nearest.`);
+    }
+
+    log.setStatus(`Querying RSGB ETCC for ${plan.squares.length} locator square(s)...`);
```
jasiek, line 180: **P2 — Keep the truncation warning in the final user-visible status.** The status set above is overwritten here in the same task, before the browser can render it, and the later insertion status also omits the warning. The accepted 300–500 km range commonly exceeds the 24-square cap (for London, 300 km is 24/30 squares and 500 km is 24/68), so the final result silently reads as complete. Please retain `plan.truncated` in the completion message, reject such searches, or switch to the full-directory endpoint.

web/js/ui/rsgb-query.js
```diff
@@ -0,0 +1,246 @@
...
+      dom.rsgbRadiusEl.focus();
+    }
+  }
+
+  function isModalOpen() {
+    return !dom.rsgbModalEl.classList.contains("hidden");
+  }
+
+  function openModal() {
+    populateOptions();
```
jasiek: **P2 — Reset the complete filter state on every open.** Rebuilding these option lists resets band and mode, but `rsgbOnlyOperationalEl.checked` retains its previous DOM property. After a user includes off-air repeaters once, every later opening continues to include them despite the PR's stated reset-to-default behavior. Please restore `checked = true` here (and reset the radius too if “all filters” is the intended contract), with a reopen test.

jasiek (commented):
I found three correctness/UX issues in the RSGB import. Details are inline.

web/js/ui/rsgb-query.js
jasiek: Yep

web/js/rsgb.js
jasiek: We're only interested in D-STAR out of all digital modes.

PR #56, opened 2026-08-16
Serial loopback test suite and adapter diagnostics page

jasiek:
Tests USB-serial adapters and the chip drivers behind them by bridging TX to RX and requiring every byte written to come straight back.

## What this adds

**`web/js/loopback-suite.js`** — the suite. Transport-agnostic: it takes anything shaped like a Web Serial `SerialPort`, so the same cases cover native `navigator.serial`, all three WebUSB chip drivers and the CDC polyfill. Cases: byte transparency (all 256 values), XON/XOFF passing through as data, writes straddling USB packet boundaries, data after an idle gap, a large single write, a clean read timeout, control lines, and reopen. Timeout budgets scale with baud so 9600 doesn't fail on the clock.

**`web/serial-test.html`** — the hardware harness. Transport picker, baud selection, per-case pass/fail table and a copyable report. It names the driver your cable actually resolved to, which is the question the page exists to answer. No Pyodide, no analytics, no manifest link — it is a bench tool that must start instantly, and it should neither mint `page_view` traffic nor offer itself for install.

**`scripts/test-loopback-harness.mjs` + `scripts/test-loopback-suite.mjs`** — the CI harness. A fake `USBDevice` whose bulk OUT feeds its bulk IN, modelling each chip's endpoint layout and FTDI's status-byte header, so `npm run test:webusb` exercises the drivers' real read/write paths with no hardware attached. Fault injectors prove each case *fails* on the defect it claims to catch — truncated writes, doubled echoes, swallowed XON/XOFF, a link that dies after one idle gap, an FTDI device sending no status header. A loopback case that cannot fail is worse than no case at all, and on real hardware a working cable makes a no-op case and a real case look identical.

## Two bugs found

- **All three chip drivers could not be reopened.** `close()` latched `_closed = true` and `open()` never cleared it, so a reopened port object rebuilt its streams around a `pull` whose `while (!isClosed())` was false on entry — and per the Streams spec a pull that resolves without enqueuing is never re-invoked. The port returned readable/writable that silently delivered nothing, with no error anywhere. Latent today: `requestPort()` builds a fresh port object per connect, so the app has no path that reopens one.
- **No global `button:disabled` rule.** Only `.left-panel` and `.editor-view-toggle` had one, so a disabled button anywhere else looked live.

## Note on what loopback cannot do

It cannot verify baud rate. Both directions clock off the same divisor, so a wrong `ch340GetDivisor()`/`ftdiConvertBaudrate()` result echoes perfectly. Divisor math stays covered by the existing per-chip unit tests against published encodings. The page says this in prose so a green run isn't over-read.

Relatedly, the control-line case skips itself on WebUSB ports: the chip drivers implement `setSignals` but not `getSignals`, so input lines are unreadable there. CHIRP never reads them either — it only ever drives DTR/RTS via `WANTS_DTR`/`WANTS_RTS` — so this costs the app nothing.

## Dependencies

None added.

## Testing

`npm test` passes: 833 tests, 0 failures. The new suite adds ~28s to `npm run test:webusb`. The page was rendered in headless Chrome to confirm it loads and lays out; running it against real silicon needs an adapter with a TX-RX jumper.

https://claude.ai/code/session_013HvrXUdFUX1ESvtC5Fp87z


web/js/loopback-suite.js
```diff
@@ -0,0 +1,520 @@
...
+      }
+      try {
+        await writer?.close();
+      } catch {
+        try {
+          writer?.releaseLock();
+        } catch {
+          // Already released.
+        }
+      }
```
jasiek, line 220: **Blocking:** the writer lock is only released when `writer.close()` *throws*. `WritableStreamDefaultWriter.close()` does not release the lock — only `releaseLock()` does — so on the success path `port.writable.locked` stays `true`.

Per the Web Serial spec, `SerialPort.close()` rejects with `InvalidStateError` when `readable` or `writable` is locked. That rejection is swallowed by the bare `catch` in `runWithOpenPort` (L470-474), so the port stays open, the next `port.open()` fails, and every remaining case is recorded as `could not open port at N baud`.

Concretely: on desktop Chrome a user picks "Native Web Serial" with 9600/38400/57600/115200 ticked. The 9600 pass runs, then the other three rates plus `sustained-throughput`, `read-timeout` and `reopen` all fail — and the report blames the adapter.

This repo's own proven teardown does the right thing; `web/js/serial.js:196-200` releases the writer lock unconditionally before `this.port?.close()`:

```js
try { this.writer?.releaseLock(); } catch { }
try { await this.port?.close(); } catch { }
```

The tests miss it because neither `createEchoPort.close()` nor the three WebUSB drivers' `close()` check stream locks — only native ports do, so the fake hardware diverges from the transport that matters most here.

Secondary risk on the same block: `await writer.close()` waits for pending writes to flush, so on a genuinely wedged adapter — the exact thing this page exists to diagnose — the run can hang indefinitely. `serial.js` deliberately never calls `writer.close()` at all, just `releaseLock()`.

web/js/loopback-suite.js
```diff
@@ -0,0 +1,520 @@
...
+    await port.open({ baudRate });
+    session = createPortSession(port);
+    session.start();
+  } catch (error) {
+    const detail = `could not open port at ${baudRate} baud: ${error?.message || error}`;
+    for (const entry of entries) {
+      results.push(makeResult({ ...entry, baudRate }, "fail", detail, ctx.now(), ctx));
+    }
+    return;
+  }
```
jasiek, line 495: A throw from `session.start()` leaks an open port. `port.open()` has already succeeded by then, but the `catch` pushes results and `return`s — there's no `finally` on this block, so `port.close()` is never called.

`start()` throws whenever `port.readable`/`port.writable` are falsy or `getReader()`/`getWriter()` hits an already-locked stream. The port then stays claimed for the life of the page: every subsequent baud rate fails, and so does a second click of "Run tests".

The recorded detail also contradicts itself — `could not open port at 9600 baud: Port exposes no readable/writable streams — is it open?` — and points the reader at the wrong thing. Worth separating "open failed" from "session setup failed", and closing the port in the latter case.

web/js/loopback-suite.js
```diff
@@ -0,0 +1,520 @@
...
+  // Cases read the baud in force off the context to size their timeouts.
+  const caseCtx = { ...ctx, baudRate };
+  let session = null;
+  try {
+    await port.open({ baudRate });
+    session = createPortSession(port);
+    session.start();
+  } catch (error) {
+    const detail = `could not open port at ${baudRate} baud: ${error?.message || error}`;
+    for (const entry of entries) {
```
jasiek: These results bypass `ctx.onCase`, so they never reach the diagnostics page's table.

`runCase()` emits `onCase({phase:"finish", ...})` for every result, and `web/js/serial-test-page.js:145-151` builds the live table *only* from those events. This path pushes straight into `results`.

So if an adapter refuses one baud rate — the scenario the comment above says this branch exists for — the user sees the status line read "0 passed, 4 failed" with an empty table above it, and only finds out why if they scroll down to the Report block.

web/js/loopback-suite.js
```diff
@@ -0,0 +1,520 @@
...
+      results.push(makeResult({ ...entry, baudRate }, "fail", detail, ctx.now(), ctx));
```
jasiek: This marks entries `fail` even when `requires()` would have skipped them. `runCase()` consults `entry.requires` first; this loop doesn't.

With the default `controlLines: false`, a port that fails to open for the `ONCE_CASES` pass reports `control-lines` as **FAIL** ("could not open port…") even though the user never claimed to have RTS/CTS jumpered and the case is inapplicable. That inflates `summary.failed` and sends someone off checking jumpers they were told were optional.

web/js/loopback-suite.js
```diff
@@ -0,0 +1,520 @@
...
+  if (baudRates.length === 0) {
+    throw new Error("runLoopbackSuite needs at least one baud rate");
+  }
+  const results = [];
+
+  for (const baudRate of baudRates) {
+    await runWithOpenPort(port, baudRate, PER_BAUD_CASES, ctx, results);
+  }
+
+  await runWithOpenPort(port, baudRates[baudRates.length - 1], ONCE_CASES, ctx, results);
```
jasiek, line 542: "The highest baud rate" (per the `ONCE_CASES` comment) is really "the last element" — `runLoopbackSuite` takes the caller's array as-is and never sorts it.

`runLoopbackSuite(port, { baudRates: [115200, 9600] })` runs `sustained-throughput` at 9600, where 16384 bytes needs a ~34 s round trip against a ~4 s timeout budget: a guaranteed spurious failure.

The page happens to sort (`serial-test-page.js:61`), but this is an exported public API and `test-loopback-suite.mjs` only ever passes sorted input. A `.slice().sort((a, b) => a - b)` at L486 would make the comment true.

web/js/serial-test-page.js
```diff
@@ -0,0 +1,253 @@
...
+
+// A header the report can be read out of context with: what was tested, on
+// what, in which browser.
+function buildReport(summary) {
+  return [
+    `Adapter: ${chosenDescription}`,
+    `Browser: ${navigator.userAgent}`,
+    `Baud rates: ${selectedBaudRates().join(", ")}`,
+    `Control lines bridged: ${dom.controlLines.checked ? "yes" : "no"}`,
+    "",
```
jasiek: The report header can misstate what was actually tested. `onRun` disables `dom.run` and `dom.choose` but not the baud checkboxes or the control-lines checkbox, and `buildReport()` reads `selectedBaudRates()` and `dom.controlLines.checked` *after* the run finishes (L198).

A run takes tens of seconds. A user who ticks 19200 mid-run gets a report claiming `Baud rates: 9600, 19200, 38400, 57600, 115200` for a run that never touched 19200 — and this report is explicitly meant for pasting into issues.

Either snapshot both values at L169 next to `baudRates` and pass them into `buildReport`, or disable the whole control group for the duration.

web/serial-test.html
```diff
@@ -0,0 +1,100 @@
+<!doctype html>
```
jasiek, line 1: Nothing links to this page. Neither `web/index.html` nor `web/about.html` references `serial-test.html`, and it isn't in the README — it has a "Back to App" link out but no route in, so it's reachable only by typing the URL.

If that's deliberate (an unlisted bench tool), worth a line in `FINDINGS.md` or the README so whoever needs it can find it.

jasiek (commented):
Code review pass over the whole diff, focused on correctness.

One finding I'd treat as blocking (**#1**): `createPortSession.close()` never releases the writer lock on the success path, which makes `port.close()` reject on native Web Serial — the page's primary transport — and cascades into spurious "could not open port" failures for every case after the first baud rate. The fake ports in the harness can't catch it because neither `createEchoPort` nor the three WebUSB drivers check stream locks on close; only native ports do.

The rest are lower severity: an open-port leak on one error path, results that bypass `onCase` and so never reach the page's table, and a couple of reporting inaccuracies.

Things I checked and cleared:
- `_closed = false` is applied consistently across all three drivers and matches each one's `cancel`/`close` latch. The `FINDINGS.md` claim that no app path reopens a port checks out — `serial.js:146` requests a fresh port per connect and `_teardown()` nulls `this.port`.
- The new global `button:disabled` rule can't regress existing controls: every later scoped `:disabled` rule uses identical values at higher specificity, and `#mem-table .channel-location-button` (1,1,0) outranks it.
- `build-dist.mjs` copies all of `web/`, so `serial-test.html` and its module graph ship and get hash-rewritten; there's no service worker precache list to update.
- `npm run test:webusb`'s two new files pass, 28/28.

web/js/loopback-suite.js
jasiek: fix this

web/js/loopback-suite.js
jasiek: fix

web/js/loopback-suite.js
jasiek: fix

web/js/loopback-suite.js
jasiek: fix

web/js/serial-test-page.js
jasiek: fix

web/serial-test.html
jasiek: leave as is

web/js/loopback-suite.js
jasiek: Already fixed in 61d1835 (pushed before this reply landed). `createPortSession.close()` now calls `writer?.releaseLock()` and never `writer.close()`, mirroring `web/js/serial.js:196-200` — so the writable is unlocked on the success path and `SerialPort.close()` no longer rejects with `InvalidStateError`, and there is no flush to hang on with a wedged adapter.

The fakes now catch this class of bug too: `createEchoPort.close()` rejects while either stream is locked, which is the spec check the old fake was missing.

web/js/loopback-suite.js
jasiek: Already fixed in 61d1835. `runWithOpenPort` tracks an `opened` flag, so a throw from `session.start()` closes the port again instead of leaving it claimed for the life of the page. The two failure modes are also worded apart now — `could not open port at N baud` vs `could not start reading from the port at N baud` — so the detail no longer contradicts itself.

web/js/loopback-suite.js
jasiek: Already fixed in 61d1835. Every result now goes through a single `recordResult()` helper that pushes to `results` *and* emits `ctx.onCase({phase:"finish", ...})`, so the open-failure path reaches the page table like any other result. No path constructs a result without announcing it.

web/js/loopback-suite.js
jasiek: Already fixed in 61d1835. `failEntries()` consults `entry.requires(port, ctx)` first and records `skip` with the requirement text rather than `fail` with the open-failure detail — so `control-lines` reports SKIP when the user never claimed the jumpers, and `summary.failed` is not inflated.

web/js/serial-test-page.js
jasiek: Already fixed in 61d1835. `onRun()` snapshots `{baudRates, controlLines}` at the start and passes it into `buildReport()`, which no longer reads the DOM. Belt and braces: `setControlsDisabled()` now locks the transport select, the control-lines checkbox and every baud checkbox for the duration, not just the two buttons.

web/js/loopback-suite.js
jasiek: Fixed in 61d1835 as well, though this thread was not marked for it. `runLoopbackSuite` now does `ctx.baudRates.slice().sort((a, b) => a - b)`, so `ONCE_CASES` picking off the end really is the highest rate and `runLoopbackSuite(port, {baudRates: [115200, 9600]})` no longer runs the 16 KB throughput case at 9600 against a budget computed for 115200.

web/serial-test.html
jasiek: Leaving the page unlinked as directed — no link added from `index.html`, `about.html` or the nav. 61d1835 does add a README paragraph recording that it is deliberately unlisted and how to reach it; say the word if you would rather that came out too.

PR #59, opened 2026-08-17
Queue bulk IN transfers on the CH340 and PL2303 drivers

jasiek:
Loopback tests were failing on a phone with a CH340G cable. Diagnosed on a
Pixel 10 (Chrome 151) over wireless adb, driving the page from a local checkout
so the drivers could be instrumented and reloaded against the real adapters.
A PL2303HX was then tested the same way and turned out to have the same defect.

## The fault

Both read paths issued one bulk IN transfer at a time, so the endpoint sat
unqueued for the whole round trip between one transfer completing and the next
being issued. The chip's RX FIFO overran in the gap and dropped bytes with
`status: "ok"` and no error at any layer.

The endpoint size sets how much margin there is, which is why the two chips fail
so differently:

| | endpoint | budget @115200 | symptom before fix |
|---|---|---|---|
| CH340G | 32 B | one transfer per 2.8 ms | loses bytes at **every** size ≥ 128; 16 KB drops ~208 B |
| PL2303HX | 64 B | one transfer per 5.6 ms | 256-byte cases pass; only the 16 KB echo fails, on 3 of 10 runs |

The originally reported CH340 failures were two 256-byte cases, which is
misleading — size has nothing to do with it. Sweeping sizes and rates showed
115200 losing bytes from 128 bytes up, while 9600 and 57600 were clean to 16 KB.
The loss is probabilistic, which is why the 16 KB case passed in the same run
that failed 256 bytes.

## The fix

Keep a pipeline of transfers queued, and give the stream a matching high-water
mark. The high-water mark is load-bearing on its own — a queue of one puts the
consumer's own work back on the USB critical path, and depth 8 without it still
dropped a packet mid-suite.

Linux has done this all along, and `ch341.c`/`pl2303.c` are where to read it:
neither sets `bulk_in_size` or a read callback, so both inherit `read_urbs[2]`
(`include/linux/usb/serial.h`) — two URBs submitted at open, each resubmitted
from its own completion callback, so the endpoint is never idle. Depth 2
suffices when the resubmit happens in interrupt context; a JS resubmit crosses
Chrome's IPC boundary and the event loop, hence 16.

Every transfer still asks for **exactly one packet**. Asking for more is the
obvious cure and is wrong on the CH340: a bulk IN transfer ends on a short
packet or on the full requested length, and that chip sends neither, so against
a 512-byte request a 32-, 64- or 96-byte reply never arrives at all — measured
as 0 bytes delivered, the transfer simply stays pending. That is exactly the
shape of a fixed-size clone block. The PL2303 does terminate multi-packet
transfers correctly, but the rule is applied there too: nothing is gained once
the queue is deep, and the downside if a chip misbehaves is a silent hang.
Tests pin the request length on both so it cannot be "optimised" later.

## Verification

On hardware, with TX bridged to RX:

- **CH340G:** 17 passed / 2 failed / 1 skipped → **19 / 0 / 1**, four consecutive runs.
- **PL2303HX:** 3 failures in 10 suite runs → **0 in 12**.

`npm test` passes (854 tests), including four new regression tests covering the
pipelining and the one-packet request length on both drivers.

## Worth knowing if you reproduce this

An isolated benchmark is *less* sensitive than the suite. A bare pump over the
raw stream showed no PL2303 loss at any size or rate, because the suite's own
per-chunk buffering is part of what starves the endpoint. The defect only
appears when a real consumer is doing real work between reads.

## Not addressed

FTDI is now the only driver still reading one transfer at a time. It has the
same 64-byte endpoint as the PL2303 and therefore the same exposure, so it is
very likely affected in the same way — left alone here purely for want of
hardware to test against, and recorded in `FINDINGS.md`.

The read-pipeline logic is now duplicated across the two drivers rather than
factored into a shared helper. That is deliberate: the CH340 was verified on
hardware before the cable was swapped out, and a refactor could not be re-tested
against it in this session. Worth extracting once both cables are to hand.

## Dependencies

None added.


jasiek:
Review finding (P2): stall recovery mixes stale and fresh reads.

In `web/js/ch340-webusb.js:369-380`, after the oldest queued `transferIn` reports `stall`, the loop clears the halt and then calls `topUp()` before inspecting the remaining pre-stall transfers. If several old requests also report `stall`, this submits fresh reads and then repeatedly calls `clearHalt()` while those newer reads are pending, risking dropped data or a wedged stream.

The WebUSB specification treats these as separate operations: `clearHalt()` issues `ClearFeature(ENDPOINT_HALT)`; it does not abort transfers already scheduled on the endpoint. Recovery should retire/abort the old transfer generation, clear the halt once, and only then rebuild the pipeline (or fail/reopen cleanly if pending reads cannot be cancelled).

The test at `scripts/test-ch340-webusb.mjs:225` injects only one stalled result, so it cannot exercise a pipeline-wide stall. The new pipeline test at line 244 also asserts only `> 1`, rather than pinning the load-bearing depth of 16 and sustained refill behavior.

Validation: all 850 local tests pass. GitHub reports no configured checks for this branch.

Model used: OpenAI Codex (GPT-5).

web/js/ch340-webusb.js
```diff
@@ -313,16 +333,50 @@ export class Ch340SerialPort {
...
+            if (!transfer) {
+              return;
+            }
+            const result = await transfer;
             if (result.status === "stall") {
+              // Never wait for the rest of the queue here: a device that does
+              // not retire its queued transfers would wedge the read path for
+              // good. Clearing the halt retires them anyway, and each comes
+              // back stalled and is skipped on its own turn round this loop.
               await device.clearHalt("in", inEndpoint);
```
jasiek, line 387: [P2] Rebuild the pipeline generation after clearing a stall

On Chromium's libusb backend, `clearHalt()` cancels every outstanding transfer on the claimed interface before it clears the endpoint. Blink exposes those cancellations as rejected `AbortError` promises. Because the remaining cancelled promises stay in `inFlight`, the next iteration shifts one, `await transfer` rejects, and the outer catch permanently errors the readable stream. I reproduced that behavior locally by having the fake reject its queued transfers from `clearHalt()`.

Retire the pre-stall generation, tolerate its cancellation results, and refill only after recovery. The duplicated PL2303 path needs the same change.

Chromium sources: [`ClearHalt` cancels interface transfers](https://chromium.googlesource.com/chromium/src/+/988ef7646e298b923c329e1d56da216912544494/services/device/usb/usb_device_handle_impl.cc#657), [Blink maps cancellation to `AbortError`](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/modules/webusb/usb_device.cc#72).

scripts/test-pl2303-webusb.mjs
```diff
@@ -212,3 +216,38 @@ test("WebUSB provider dispatches Prolific devices to the PL2303 driver", async (
...
+  // loopback suite's 16 KB echo at 115200 lost bytes on 3 of 10 runs with one
+  // transfer in flight, and none of 12 once queued.
+  const { device, transferInCalls } = makeFakeDevice({ descriptor: HX_DESCRIPTOR });
+  const port = new Pl2303SerialPort(device);
+  await port.open({ baudRate: 115200 });
+  // The stream pulls as soon as it is constructed, which primes the queue.
+  await new Promise((resolve) => setTimeout(resolve, 0));
+
+  assert.ok(
+    transferInCalls.length > 1,
```
jasiek: [P3] Pin the load-bearing depth and sustained refill behavior here too

This assertion only proves that more than one initial transfer was submitted. It still passes if the depth is reduced to two or if the `CountQueuingStrategy` is removed, even though the PR describes the matching high-water mark as load-bearing. Complete enough fake transfers without attaching a reader and assert the expected replenishment count, mirroring the fix requested in the existing CH340 test thread.

jasiek (commented):
Two findings from the follow-up review.

web/js/ch340-webusb.js
jasiek: Confirmed and fixed in d55224c — thank you, this was a real defect and my comment asserting the opposite was wrong.

I checked the Chromium mapping you linked: `CheckFatalTransferStatus` rejects `UsbTransferStatus::CANCELLED` with `kAbortError`, so a cancelled transfer never reaches the `result.status` check at all. The old comment claimed the queued transfers "come back stalled and are skipped on their own turn round this loop" — they come back *rejected*, hit the outer catch, and `controller.error()` kills the stream for good.

The queue is now dropped before the halt is cleared, and refilled only after recovery. Nothing is awaited there, so the hang risk that motivated the original shape is still avoided. Applied to the PL2303 path too.

The more useful part of this: `test-loopback-harness.mjs` had modelled `clearHalt` as leaving outstanding transfers alive, which is exactly why the existing stall-recovery cases passed on the broken driver. The fake now cancels its waiting transfers — and importantly takes no wire bytes with them, since a cancelled transfer leaves the data in the device FIFO for the next one. With that model corrected, `the ch340 driver recovers from a stalled bulk IN endpoint` and its PL2303 twin fail on the unfixed driver, so the end-to-end cases carry this now rather than only the unit tests.

scripts/test-ch340-webusb.mjs
```diff
@@ -238,3 +240,41 @@ test("Ch340SerialPort read path clears a stalled IN endpoint and passes raw byte
...
+  const port = new Ch340SerialPort(device);
+  await port.open({ baudRate: 115200 });
+  // The stream pulls as soon as it is constructed, which primes the queue.
+  await new Promise((resolve) => setTimeout(resolve, 0));
+
+  assert.ok(
+    transferInCalls.length > 1,
+    `expected several transfers queued at once, saw ${transferInCalls.length}`,
+  );
+});
```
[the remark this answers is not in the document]
jasiek: Fixed in d55224c. You are right that `> 1` pinned neither the depth nor the queuing strategy.

Both tests now complete 16 packets with no reader attached and assert 31 transfers — 16 queued up front plus one replenished per pull for the 15 pulls after the first. That count only holds at the real depth with the high-water mark in place: at depth 2 it is 17, and with the default queue of 1 the stream stops after a single pull at 16.

Each assertion was checked against the defect it claims to catch. Reverting the clearHalt fix, changing `READ_PIPELINE_DEPTH` to 2, and removing the `CountQueuingStrategy` each fail their own test and no others.

scripts/test-pl2303-webusb.mjs
jasiek: Fixed in d55224c, mirroring the CH340 test: 16 packets completed with no reader attached, asserting 31 transfers, which pins the depth and the `CountQueuingStrategy` together.

The PL2303 fake also gained a `clearHalt` (it had none, so the stall path was entirely uncovered here) and a matching stall-recovery test, verified to fail without the driver fix.

jasiek:
Re-verified on hardware at d55224c, after the review round.

**CH340G** (`0x1a86:0x7523`, chip version 0x30, 32-byte endpoint): six consecutive full loopback runs, **19 passed / 0 failed / 1 skipped** each.
**PL2303HX** (`0x067b:0x2303`, 64-byte endpoint): six consecutive runs, same result, reported earlier.

So both cables are confirmed against the current head, not just against the original fix.

Worth being explicit about what this does *and does not* cover: the only CH340 change in the review round was one line inside the `if (result.status === "stall")` branch, and no stall occurred in any of roughly twenty hardware runs across both cables — a healthy adapter does not halt its endpoint. Hardware therefore confirms no regression on the normal read path; it cannot exercise the stall fix at all. That path is covered by `test-ch340-webusb.mjs`, `test-pl2303-webusb.mjs` and the corrected `test-loopback-harness.mjs`, each checked to fail on the unfixed driver.

FTDI remains untested and is the one known gap, recorded in `FINDINGS.md`.

PR #63, opened 2026-08-17
Release stuck PTT by opening previously-granted ports on sight (#60)

### LINKED ISSUE #60: When some radios are connected, they start transmitting right away
I suspect PTT is engaged on some radio models - notably the Baofeng UV-5R. Perhaps there's a way to disengage PTT before we start doing anything.

jasiek:
Fixes #60.

## Diagnosis

On the Baofeng UV-5R (and other radios with Kenwood-style two-pin cables), PTT is engaged by pulling the mic line to ground — and the programming cable drives that same line from the USB adapter's TX output. An unconfigured adapter chip rests TX low from the moment it gets USB power, which is electrically identical to a held PTT, so the radio transmits an open carrier until *any* program opens the port and the UART settles at idle-high. Verified on hardware: the radio keys the instant the cable's USB end is plugged in (before any browser code runs, on any driver), and opening the port with `screen` releases it immediately.

The browser is not allowed to touch a serial device before the user grants access, so this cannot be prevented — but it can be shortened to near zero.

## Change

- **`web/js/ptt-guard.js`** (new): opens ("parks") every previously-granted `navigator.serial` port on page load, and newly plugged ones via the `connect` event, so an affected radio stops transmitting within a fraction of a second instead of waiting for a Connect click. Parked ports stay open so TX stays driven; guard activity is logged to Debug Output.
- **`web/js/serial.js`**: two optional hooks on `BrowserSerialBridge` — `onBeforeOpen` (the guard releases parked ports before the port picker shows, otherwise the chosen port would fail to open) and `onAfterTeardown` (the guard re-parks after a session ends). Guard failures never block connecting.
- **`web/index.html` / `web/styles.css`**: a short hint under the connect buttons for first-time visitors, who have no granted port for the guard to protect yet.
- **`scripts/test-ptt-guard.mjs`**: unit tests for parking, connect/disconnect events, suspend/resume, failure tolerance, and the bridge hook ordering.
- **`FINDINGS.md`**: records the mechanism.

Scope note: native Web Serial only. Parking a WebUSB adapter would require instantiating a chip-specific driver outside a user-chosen connection; can be a follow-up if Android users report the same symptom.

## Dependencies

No new dependencies.

## Validation

`npm test` — all suites pass (866 tests, including the 7 new ones).


jasiek:
This doesn't fix the problem completely, so I'm not going to merge this in.

PR #65, opened 2026-08-17
Show repeater location on a static OSM map (#57)

### LINKED ISSUE #57: When viewing a repeater, display it on a map
Suppose you've queried the API for a list of repeaters. For those repeaters which have a location (lat/long), provide a facility to display a map which provides context for where the repeater is. The map needs to be OSM based. Repeater location is indicated by a red dot and map is to be centered on the repeater. Map can be static (no interaction is expected).

## Behavior on desktop
- Hover over the Location column - if a location is available for a repeater, display a map as a tooltip (only visible when hovering over the channel Location cell). Map should be no bigger than the sidebar on the left. Map contains coordinates in decimal form as a centered header on the top, and the rest is occupied by the map. 

## Behavior on mobile
- Click on the Location column - if a location is available for a repeater display a modal map over the rest of the app. Map should be a square and contain coordinates in decimal form like on the desktop, the map, and additionally, there needs to be an X button in the top right to dismiss the map. Clicking outside of the map widget also dismisses it.

jasiek:
Closes #57.

Channels imported from a repeater directory now offer a static OSM context map centered on the repeater: red dot, decimal coordinates as a header, OSM attribution, no interaction.

- **Desktop** (hover-capable devices): hovering the Location cell shows the map as a tooltip beside the cell, sized to the 280px sidebar. Rows that have a map mark their channel number with a dotted underline.
- **Mobile** (`hover: none`): tapping the Location cell opens a square modal, dismissed by its ✕ button, a tap outside the card, or Escape.

How it works:

- `parsePrzemiennikiXml` reads each repeater's `<location>` latitude/longitude (present on both przemienniki and repeaterbook proxy responses; ~90% of a Poland 2m sample carried coordinates). Imported rows keep them as a `__geo` sidecar (`web/js/row-geo.js`) under a non-header key, so TSV/CSV/upload serialization never sees it.
- `web/js/staticmap.js` holds the pure Web-Mercator tile planning; `web/js/ui/repeater-map.js` owns both surfaces. Tiles load with `crossOrigin="anonymous"` because the app's `COEP: require-corp` (needed by Pyodide) blocks plain cross-origin images — caught in live verification and recorded in FINDINGS.md.
- A `repeater_map_shown` analytics event fires when a map stays visible past one second (a pointer passing through reports nothing). No parameters beyond the automatic display mode; coordinates are never sent. Timer cancellation is surface-scoped so the slight table scroll from the tap that opens the modal cannot kill the modal's pending dwell.

Testing: new `scripts/test-staticmap.mjs` (projection pinned to a known OSM tile, tile-plan coverage/centering/antimeridian wrap, coordinate formatting, geo-sidecar validation and serialization exclusion) added to `test:channels`; full suite green (4 suites, 0 failures). Both surfaces plus the dwell-event timing verified end-to-end in headless Chrome against a live dev server with a captured przemienniki response.

New dependencies: none. Map tiles are fetched at runtime from tile.openstreetmap.org with the required attribution shown on every map.

https://claude.ai/code/session_01CcogzZXGsy5RS2pszVKqW9


web/js/datasources.js
```diff
@@ -333,6 +337,7 @@ export function buildPrzemiennikiRows(repeaters, { createBlankRow, setRowValue,
     if (mappedMode) {
       setRowValue(row, "Mode", mappedMode);
     }
+    setRowGeo(row, repeater.latitude, repeater.longitude);
```
jasiek, line 340: **P1 — RSGB is not wired into the map metadata path.** This call only runs in `buildPrzemiennikiRows`, which covers Przemienniki and RepeaterBook. RSGB uses `buildRsgbRows` in `web/js/rsgb.js`; although `filterRsgbRecords` already decodes each locator to `box.latitude` / `box.longitude`, those coordinates are discarded and no `setRowGeo` call is made. Consequently RSGB rows never get `.has-geo` and neither map surface can open. Please carry the decoded locator centre into the filtered entry and attach it to the built row, with a regression test asserting `rowGeo()` for an RSGB row.

web/js/ui/repeater-map.js
```diff
@@ -0,0 +1,198 @@
...
+      img.style.top = `${tile.top}px`;
+      img.src = osmTileUrl(tile);
+      canvasEl.appendChild(img);
+    }
+    const marker = document.createElement("div");
+    marker.className = "repeater-map-marker";
+    canvasEl.appendChild(marker);
+    const attribution = document.createElement("div");
+    attribution.className = "repeater-map-attribution";
+    attribution.textContent = OSM_ATTRIBUTION;
```
jasiek: **P1 — the OSM attribution does not expose the licence details required by the tile provider policy.** This creates only plain text, and there is no OpenStreetMap copyright/ODbL link elsewhere in the app. The OSM tile policy requires attribution following its guidelines, which require a way to access origin and licence information (normally by linking “OpenStreetMap” to `https://www.openstreetmap.org/copyright`). The desktop tooltip is also `pointer-events: none`, so any link placed inside it would be unusable. Please add an accessible copyright/licence link or equivalent complete licence information. References: https://operations.osmfoundation.org/policies/tiles/ and https://osmfoundation.org/wiki/Licence/Attribution_Guidelines

web/js/ui/repeater-map.js
```diff
@@ -0,0 +1,198 @@
...
+      if (button && !button.contains(event.relatedTarget)) {
+        hideTooltip();
+      }
+    });
+    // Scrolling recycles row elements under the cursor; the anchor cell may
+    // now show a different channel, so the tooltip must not linger.
+    dom.tableScrollEl.addEventListener("scroll", hideTooltip, { passive: true });
+
+    dom.tableBody.addEventListener("click", (event) => {
+      if (hoverCapable()) {
```
jasiek, line 239: **P2 — keyboard-generated clicks are discarded on desktop.** A focused geo Location button activated with Enter or Space reaches this handler, sees `hoverCapable()`, and returns. Because the tooltip is opened only from `mouseover`, keyboard users on hover-capable devices have no way to display the map. Please add focus/keyboard activation and an accessible indication that a map is available, with a regression test for the desktop keyboard path.

web/js/ui/repeater-map.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  // --- Mobile modal ---------------------------------------------------------
+
+  function openModal(geo) {
+    dom.repeaterMapModalCoordsEl.textContent = formatCoordinates(geo.latitude, geo.longitude);
+    // Show first: the canvas has no layout width while the overlay is hidden.
+    // The map is square, sized to the modal card's width (issue #57), which
+    // the stylesheet caps well inside the viewport.
+    dom.repeaterMapModalEl.classList.remove("hidden");
```
jasiek, line 179: **P2 — the modal has no focus lifecycle.** Removing `hidden` does not move focus into the dialog, constrain focus while it is open, or restore focus to the Location button after dismissal. On touch with a screen reader, focus therefore remains on the obscured table and can continue through background controls despite `aria-modal="true"`. Please focus the close button or dialog on open, keep focus within the modal, restore the trigger on close, and cover close-button, backdrop, and Escape dismissal in tests.

FINDINGS.md
```diff
@@ -68,6 +68,8 @@ CHIRP `b7ae1b6`.
...
+- **coep-blocks-plain-cross-origin-images** (2026-08-17): the app is served with `Cross-Origin-Embedder-Policy: require-corp` (both `scripts/dev-server.mjs` and production need it for Pyodide's cross-origin isolation), which silently blocks every plain cross-origin `<img>` — a repeater-map OSM tile loads fine from `curl` yet fails in the page with `ERR_BLOCKED_BY_RESPONSE.NotSameOriginAfterDefaultedToSameOriginByCoep` and shows only as a broken icon. The out: `tile.openstreetmap.org` sends `Access-Control-Allow-Origin: *`, so `img.crossOrigin = "anonymous"` turns the load into a CORS request that satisfies COEP (`web/js/ui/repeater-map.js`). Any future cross-origin asset needs either CORS mode + an ACAO header or a `Cross-Origin-Resource-Policy: cross-origin` response header on the upstream.
```
jasiek: **P3 — this production-COEP statement contradicts the existing hosting finding below.** `FINDINGS.md` line 98 correctly records that only `scripts/dev-server.mjs` sets the isolation headers and GitHub Pages cannot send them. Please describe this as a local-development constraint rather than saying production needs or serves COEP; otherwise future cross-origin debugging will start from the wrong deployment model.

web/js/datasources.js
jasiek: Add support for this source.

web/js/ui/repeater-map.js
jasiek: Add this attribution in the lower part of the map, in the bottom of the widget and make it clickable.

web/js/ui/repeater-map.js
jasiek: Leave as-is.

web/js/ui/repeater-map.js
jasiek: Keep focus on close, but screen reader support isn't critical for us.

FINDINGS.md
jasiek: Indicate that production cannot serve COEP at the moment.

PR #67, opened 2026-08-19
Drive Silicon Labs CP2102 cables over WebUSB

jasiek:
Adds a fourth WebUSB chip driver so CP210x programming cables work on Chrome for Android, where native Web Serial cannot drive them. CP2102 was previously named in the UI as the example of an unsupported chip; that string is gone.

## What it does

`web/js/cp2102-webusb.js` exposes a `Cp2102SerialPort` with the same Web Serial port surface as the FTDI, PL2303 and CH340 drivers, so `BrowserSerialBridge` uses it unchanged. The chooser filters on the Silicon Labs vendor id and dispatches to it.

Ported from the Linux `drivers/usb/serial/cp210x.c`:

- **Init sequence**: part number (`VENDOR_SPECIFIC`/`GET_PARTNUM`) → `IFC_ENABLE` → `SET_BAUDRATE` → `SET_LINE_CTL` (8N1) → `SET_FLOW` → `SET_MHS` with both control lines low. `close()` does `PURGE` then `IFC_ENABLE(disable)`.
- **Baud rate** is quantised host-side as the kernel does — AN205 Table 1 for CP2101/2/3, `div = round(48e6 / (2 × prescale × request))` from the CP2104 on — so the rate reported is the rate on the wire. Every rate CHIRP uses is an exact table entry, so nothing shifts on the radio path; the quantiser earns its place by clamping (AN205 leaves results above ~1.05 Mbaud undefined).
- **Flow control** is written explicitly at every open: hardware handshaking (CTS/DSR/DCD) and XON/XOFF off, DTR/RTS left to `setSignals`. `IFC_ENABLE` does not reset that block, and XON/XOFF would eat 0x11/0x13 out of a clone image with no error anywhere. Read-modify-write like the kernel, so the chip's own FIFO thresholds survive.
- **Requests are addressed to the claimed interface** (`bmRequestType 0x41/0xc1`, `wIndex` = interface number), which is where this family differs from the other three drivers' device-recipient requests. Hardcoding zero works on a single-UART CP2102 and addresses the wrong port on a CP2105/CP2108.

Carries over what the earlier drivers learned the hard way: a full pipeline of bulk IN transfers with a matching high-water mark, exactly one packet per transfer, and the pre-stall queue retired before `clearHalt`.

Unlike the other three it also implements `getSignals`, since `GET_MDMSTS` reports the input control lines — so the loopback page's control-line case can run on the WebUSB transport for the first time.

## Policy on failures

Informational *reads* are tolerated (a part number the chip will not report downgrades to unknown, as in the kernel; an unanswered `GET_FLOW` yields a constructed block). Config *writes* fail the open with a message naming the request, rather than continuing with unknown line settings — the kernel merely logs those, but a silently mis-configured line corrupts a clone instead of failing it. Worth relaxing later with evidence from real hardware.

## Testing

- `scripts/test-cp2102-webusb.mjs` — 19 cases: baud quantisation and clamping, vendor dispatch and chooser filter, the init sequence byte for byte, interface addressing (opened against interface 1), a part that will not report its part number, the handshake-clearing block, endpoint selection, `setSignals` write-mask semantics, `getSignals` decoding, stall recovery, pipeline depth, one-packet transfers, and close.
- cp2102 joins `CHIP_NAMES` in the shared loopback harness, so it runs the real suite against a fake USBDevice whose bulk OUT feeds its bulk IN.
- Each assertion was mutation-checked: shallower pipeline depth, multi-packet requests, a hardcoded interface index, skipped or incomplete `SET_FLOW`, dropped `clearHalt` recovery, always-write `setSignals`, and a missing `PURGE` each fail their own test and no others.
- Full suite green: 895 tests.

**Not yet run against real hardware** — hardware verification on Android is the next step.

## New dependencies

None.

## Docs

README's WebUSB section and architecture list updated; four new entries in FINDINGS.md (interface-addressed requests, why a vendor-wide filter is right here where it was wrong for WCH, the flow-control block nothing resets, host-side baud quantisation), and the `getSignals` and "all three drivers" entries revised.


web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+    let flowReplace = block.getUint32(4, true);
+
+    controlHandshake &= ~(
+      SERIAL_CTS_HANDSHAKE | SERIAL_DSR_HANDSHAKE
+      | SERIAL_DCD_HANDSHAKE | SERIAL_DSR_SENSITIVITY | SERIAL_DTR_MASK
+    );
+    if (this._dtr) {
+      controlHandshake |= SERIAL_DTR_ACTIVE;
+    }
+    flowReplace &= ~(SERIAL_RTS_MASK | SERIAL_AUTO_TRANSMIT | SERIAL_AUTO_RECEIVE);
```
jasiek: **[P1] Clear the input-byte transformation bits.** This preserves `SERIAL_NULL_STRIPPING` (bit 3), which AN571 defines as discarding every received NUL, and also preserves error/break character insertion. Since this flow block survives sessions, a prior configuration can silently shorten or expand binary clone data even though XON/XOFF is disabled here. Please clear bits 2–4 (and reserved bits as required) and add a test seeded with bit 3 set. https://www.silabs.com/documents/public/application-notes/AN571.pdf

web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+        ))
+        : null;
+    } catch {
+      block = null;
+    }
+    if (!block) {
+      block = new DataView(new ArrayBuffer(FLOW_CTL_SIZE));
+      // The SiLabs defaults for the FIFO thresholds the chip only uses when
+      // XON/XOFF is on; written so a constructed block is not all zeroes.
+      block.setUint32(8, 0x80, true);
```
jasiek: **[P2] Keep both XON/XOFF limits zero when software flow control is disabled.** CP2102N A01 erratum E104 reads the first byte of `ulXonLimit` as `ulFlowReplace`; `0x80` therefore selects RTS flow control on affected firmware, which can make `SET_MHS` fail or gate TX despite clearing the real flow word. This also applies when preserving a nonzero block returned by `GET_FLOW`. https://www.silabs.com/documents/public/errata/cp2102n-errata.pdf

web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+    if (!this._inEndpoint || !this._outEndpoint) {
+      throw new Error(
+        `CP2102: bulk IN/OUT endpoints not found on interface ${this._interfaceNumber}`,
+      );
+    }
+
+    this.partNumber = await this._readPartNumber();
+    // Enabling the interface also clears event-insertion mode, so whatever a
+    // previous driver left the chip in, the bulk IN stream is raw payload.
+    await this._controlOut(REQ_IFC_ENABLE, UART_ENABLE);
```
jasiek, line 509: **[P1] `UART_ENABLE` does not clear event-insertion mode.** AN571 says event mode is cleared by USB reset or `IFC_ENABLE(UART_DISABLE)`, not by enable. If a previous session left it active, bulk IN contains escaped status/event bytes and raw clone data is corrupted. Please explicitly disable the interface (or otherwise disable embedded events) before enabling it, and pin that sequence in a test. https://www.silabs.com/documents/public/application-notes/AN571.pdf

web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+    // critical path: nothing is re-queued on the endpoint until it has taken
+    // the previous packet. Buffering a pipeline's worth lets pull keep cycling
+    // transfers while the consumer is busy.
+    }, new CountQueuingStrategy({ highWaterMark: READ_PIPELINE_DEPTH }));
+
+    this.writable = new WritableStream({
+      write: async (chunk) => {
+        await device.transferOut(outEndpoint, chunk);
+      },
+    });
```
jasiek, line 711: **[P1] Check the bulk-OUT result.** WebUSB resolves a stalled transfer as a `USBOutTransferResult` with `status: "stall"` rather than rejecting, and it also reports `bytesWritten`. Ignoring both makes `writer.write()` report success after losing a protocol block. Please validate the status and byte count, then either recover/retry or throw so the upload aborts visibly. https://wicg.github.io/webusb/#usbouttransferresult

web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+      this._dtr = Boolean(dtr);
+    }
+    if (rts !== undefined) {
+      control |= CONTROL_WRITE_RTS;
+      if (rts) {
+        control |= CONTROL_RTS;
+      }
+      this._rts = Boolean(rts);
+    }
+    await this._controlOut(REQ_SET_MHS, control);
```
jasiek, line 429: **[P2] Commit the DTR/RTS shadow state only after this request succeeds.** The fields are currently updated before awaiting `SET_MHS`; after a failed request, an identical retry is treated as a no-op even though the hardware never changed. A fail-then-retry test would catch this.

web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+    // Clear both FIFOs before disabling the UART — the CP2108 occasionally
+    // hangs without it — then disable the interface, which also takes the chip
+    // out of event-insertion mode if anything ever turned it on.
+    try {
+      await this._controlOut(REQ_PURGE, PURGE_ALL);
+    } catch {
+      // Device may already be gone.
+    }
+    try {
+      await this._controlOut(REQ_IFC_ENABLE, UART_DISABLE);
```
jasiek: **[P2] Drive DTR and RTS low before disabling the UART.** CP2102N A01 erratum E106 says closing the port does not reset these outputs and requires the host to deactivate them. Without a best-effort `SET_MHS` here, disconnect can leave a radio or programming circuit held until unplug. https://www.silabs.com/documents/public/errata/cp2102n-errata.pdf

web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+// chip's RX FIFO overruns in that gap — silently, with `status: "ok"` and no
+// error at any layer. This chip's 64-byte bulk IN endpoint gives twice the
+// slack of the CH340's 32-byte one, which changes the margin, not the defect.
+const READ_PIPELINE_DEPTH = 16;
+
+export function isCp2102Device(device) {
+  if (!device) {
+    return false;
+  }
+  return Number(device.vendorId) === CP210X_VENDOR_ID;
```
jasiek: **[P2] A Silicon Labs VID is not enough to identify an eligible single-UART device.** This currently routes CP2102C (standard CDC) ahead of the CDC fallback, and also accepts non-UART products such as CP2130 (`10c4:87a0`). Keep the vendor-wide chooser if OEM PIDs matter, but dispatch CDC interfaces first and reject known/nonmatching products before issuing CP210x configuration writes. CP2102C should still be supported through the normal CDC path. https://www.silabs.com/documents/public/data-sheets/cp2102c-datasheet.pdf

web/js/cp2102-webusb.js
```diff
@@ -0,0 +1,560 @@
...
+      throw new Error(`CP2102: could not open USB device: ${error?.message || error}`);
+    }
+    if (!this.device.configuration) {
+      await this.device.selectConfiguration(1);
+    }
+
+    const iface = this.device.configuration.interfaces[0];
+    this._interfaceNumber = iface.interfaceNumber;
+    try {
+      await this.device.claimInterface(this._interfaceNumber);
```
jasiek, line 472: **[P3] Roll back a partially completed open.** Failures after `device.open()` or `claimInterface()` leave the device/interface owned. The main bridge happens to tear down after its open failure, but the loopback helper only calls `close()` when `open()` completed, so one rejected initialization can poison subsequent retries. Please make `open()` transactional with best-effort release/close.

README.md
```diff
@@ -66,13 +66,18 @@ dispatched to a chip-specific driver:
   clone-silicon prescaler quirk. Enumeration and baud/line configuration are
   verified against a real adapter; a full clone against a radio has not been
   exercised yet, so the bulk read path is still untested.
+- **Silicon Labs CP2102** adapters (and the rest of the CP210x family) use a
+  built-in CP2102-over-WebUSB driver that reads the part number to pick the
+  right baud generator, disables every hardware and software handshake the chip
+  can do on its own, and is the one WebUSB driver that can also *read* the input
+  control lines. Not yet exercised against real hardware.
```
jasiek: This is now stale: the CP2102 WebUSB path has been exercised successfully on a real hardware device. While updating it, I suggest naming the scope as single-UART CP2101/2/3/4/9/2N vendor-protocol devices, with CP2102C handled separately through CDC.

jasiek (commented):
Reviewed with the intended scope clarified as all single-UART Silicon Labs bridges: CP2101/2/3/4/9/2N through the vendor protocol, while CP2102C should use its standard CDC path. The completed real-device WebUSB test covers the CP2102 happy path, and the full local suite passes (895 tests). I found the sticky-state and failure-path issues noted inline; I omitted the earlier CP2105/CP2108 multi-interface concern because those devices are outside this scope.

web/js/cp2102-webusb.js
jasiek: Address this (CP210* support)

web/js/cp2102-webusb.js
jasiek: No radios require software flow control support - so by default it should be disabled.

web/js/cp2102-webusb.js
jasiek: Fix.

web/js/cp2102-webusb.js
jasiek: Validate this.

web/js/cp2102-webusb.js
jasiek: Fix this.

web/js/cp2102-webusb.js
jasiek: Fix this.

web/js/cp2102-webusb.js
jasiek: We only want to support single-UART devices, so we should probably check against a product list.

web/js/cp2102-webusb.js
jasiek: Fix this.

README.md
jasiek: Update these notes.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1. `ulFlowReplace` bits 2-4 — error char, NUL stripping, break char — are now cleared alongside the RTS and XON/XOFF bits. The test seeds a chip reporting all of them on (flowReplace `0x9f`) and asserts the written word is zero; removing the mask fails it.

Left bit 31 (`XOFF_CONTINUE`) as the chip reports it, matching the kernel, which touches neither it nor the other reserved bits. Say the word if you want that zeroed too.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1. Both limits now go out as zero unconditionally — including when a `GET_FLOW` block is being preserved, which is the case the erratum makes dangerous. With software flow control off they are unused, so there is nothing to weigh against E104.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1 — you are right, and my code comment asserted the opposite. `open()` now sends `IFC_ENABLE(UART_DISABLE)` before `IFC_ENABLE(UART_ENABLE)`, and the test pins both requests in order; dropping the disable fails it.

Chose the interface disable over `EMBED_EVENTS(0)` because it is the kernel-sanctioned path and cannot stall on a counterfeit part that never implemented event mode. FINDINGS carried the same wrong claim and is corrected.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1. The write path now checks `status` and resumes from `bytesWritten`: a short transfer continues from where it stopped, a stall clears the halt and throws naming the offset, and a transfer that reports zero progress throws rather than spinning. Three tests cover it.

Worth flagging: the FTDI, PL2303 and CH340 drivers all still have the unchecked `await device.transferOut(...)` form. I left them alone to keep this PR in review scope — all three are hardware-verified as they stand — and recorded it in FINDINGS as a follow-up.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1. The shadow is assigned after `await this._controlOut(...)` returns. The fail-then-retry test stalls `SET_MHS`, asserts the rejection, then repeats the identical call and asserts the request actually goes out; moving the assignment back above the await fails exactly that test.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1. `close()` now writes both lines low unconditionally before `PURGE` and the UART disable, best-effort so a device already gone does not turn teardown into an error.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1, taking the option you left open: the chooser filter stays vendor-wide so the ~150 OEM cables with custom product ids remain visible, and dispatch does the narrowing.

`isCp2102Device()` now rejects a product list — CP2105/CP2108 (`ea70`/`ea71`/`ea7a`/`ea7b`), CP2110 (`ea80`), CP2112 (`ea90`), CP2130 (`87a0`) — and separately declines any device whose descriptors declare a CDC interface class, which is what catches the CP2102C: it shares product ids with its siblings, so only the class code separates them, and the descriptors are readable before `open()`. Declining makes it fall through to the polyfill, so CP2102C is supported through the normal CDC path.

I went with a blocklist rather than an allowlist because an allowlist is what would hide the OEM cables. If you would rather have an allowlist and accept that cost, it is a one-line change.

web/js/cp2102-webusb.js
jasiek: Fixed in 02b7ce1. Initialization moved into `_initialize()`, called inside a try in `open()`; any failure past `device.open()` runs a best-effort `releaseInterface`/`close` before rethrowing. The test stalls the first configuration write, asserts the interface is released and the device closed, then reopens successfully against a healthy chip — without the rollback the retry fails at `claimInterface` with the misleading "another driver may already control it".

README.md
jasiek: Updated in 02b7ce1: the note now says verified on Android Chrome against a real CP2102 cable, and names the scope as the single-UART vendor-protocol parts (CP2101/2/3/4/9, CP2102N), with CP2102C handled through the CDC path and CP2105/CP2108 out of scope. The module header and the chooser comment say the same thing.

PR #69, opened 2026-08-20
Unify the repeater-query modals into one component-driven UI

jasiek:
## Summary

Replaces the two near-duplicate query modals (przemienniki/RepeaterBook in `repeater-query.js`, RSGB in `rsgb-query.js`) with a single modal shell assembled from reusable field components, plus the UX changes layered on top.

### Architecture
- **`web/js/ui/query-fields.js`** — field components (select, fixed value, checkbox group, boolean checkbox, number, position cluster with two-way Maidenhead locator sync) that build their own DOM from config. Behaviour is identical for every source; only labels, options and defaults differ.
- **`web/js/ui/repeater-sources.js`** — per-source configs: fields, option loading (cached `/meta` dictionary for przemienniki/RepeaterBook; static constants for RSGB) and `runQuery`. The flows stay per-source: the remote directories filter server-side via the CORS proxy, RSGB fans out over locator squares and filters client-side (still bare GETs, still available when the proxy is disabled).
- **`web/js/ui/repeater-query.js`** — the shared shell: one modal in `index.html`, form rebuilt per open (which is the reset policy: filters return to per-source defaults, position persists across opens and source switches). `rsgb-query.js` is deleted; `dom.js` drops 24 field ids for 6 shell ids.

### UX changes
- Geolocate 🛰️ moved next to the Locator field, with a new 🗑️ button that clears all location fields (including the persisted position).
- Band/mode labels unified to the dictionary casing across sources (70cm/2m/…, fm/dstar); values sent stay each API's own codes.
- RSGB shows its unsupported modes (dmr, p25, nxdn, m17) as disabled checkboxes with a "Only analogue modes and dstar are supported fully" tooltip.
- Every source now opens with 2m + 70cm on fm pre-selected.
- Behavioural corrections riding along: przemienniki filters no longer silently persist across opens; the `repeater_import` `located` flag now means a validated position; out-of-range coordinates are no longer sent upstream; RepeaterBook geolocation errors are no longer labelled "Przemienniki"; the modal focuses its first focusable control.

### Tests
- New `test-ui-query-fields.mjs` unit-tests each component in isolation; new `test-ui-repeater-modal.mjs` drives the unified modal directly (ports the whole RSGB suite and adds the first coverage of the przemienniki query URL parameters). Net −400 lines with more coverage.
- Full suite green; all three modals verified rendering correctly in the real app via headless Chrome against `npm run dev`.

## New dependencies

None.


web/js/ui/repeater-query.js
```diff
@@ -23,292 +25,116 @@ function resolveRepeaterApiBase() {
...
     ctx.table.setMenuOpen(false);
-    log.setStatus(`Loading ${source.label} query options...`);
-    await ensureDictionaryLoaded();
-    // The coordinate fields are plain DOM properties that survive a close, so
-    // bring the locator back in step with whatever they still hold.
-    refreshLocatorFromCoords();
+    let loadedOptions = null;
+    if (source.loadOptions) {
+      log.setStatus(`Loading ${source.label} query options...`);
+      loadedOptions = await source.loadOptions();
```
jasiek, line 126: `openModal()` can race across sources: select a source whose `/meta` request is slow, then another whose request is fast, and the second form can render before the first completion replaces it. There is no generation check around this `await`, so completion order wins over click order and can discard edits to the newer form. Please increment/capture an open generation before loading and ignore stale completions (or abort the previous load) so the latest selection always wins.

web/js/ui/repeater-query.js
```diff
@@ -326,63 +152,59 @@ export function createRepeaterQuery(ctx) {
...
+    dom.repeaterQueryFormEl.addEventListener("submit", async (event) => {
       event.preventDefault();
       try {
-        await runQuery();
+        if (!state.currentHeaders.length) {
+          log.setStatus("No channel schema loaded yet.");
+          setModalOpen(false);
+          return;
+        }
+        await activeSource.runQuery(collectValues());
```
jasiek, line 204: This async handler captures neither the source nor the modal generation. Repro: start a slow RSGB query, close the modal, then open RepeaterBook; when the RSGB request finishes, this closes the RepeaterBook modal. If the old request fails, the catch below labels it using the current `activeSource`. Please capture `{ source, values, modalGeneration }` before awaiting, use the captured source for errors, and only close the modal if that generation is still current. Disabling/deduplicating submission while a request is active would also prevent double submits.

web/js/ui/repeater-sources.js
```diff
@@ -0,0 +1,345 @@
...
+          name: "mode",
+          options: [
+            ...RSGB_MODES.map((mode) => ({
+              value: mode.value,
+              label: RSGB_MODE_LABELS[mode.value] || mode.label,
+              title: `${mode.label} (${mode.value})`,
+            })),
+            ...RSGB_UNSUPPORTED_MODES.map((mode) => ({
+              ...mode,
+              disabled: true,
```
jasiek, line 230: These disabled modes are not actually excluded from an unfiltered query. If the user unticks both FM and D-STAR, `values.modes` is empty, and `filterRsgbRecords()` deliberately treats an empty mode set as “any.” DMR/P25/NXDN/M17 records can then pass the filter and may be built when the selected radio advertises those modes, despite this UI presenting them as unavailable. Please either enforce the supported-mode set when nothing is selected, omit these disabled controls, or provide an explicit and accurately described “Any mode” policy.

jasiek:
Overall design assessment: the shared modal shell and shared position behavior are good directions, but the generic field-description layer is more abstraction than the current sources require. All three forms have the same six rows; only labels, options, defaults, and fixed-versus-select country presentation differ. A single static form configured per source, plus separate remote-directory and RSGB query adapters, would preserve the useful consolidation while avoiding the 303-line field DSL, dynamic factory dispatch, and bespoke DOM-testing surface. Production query JavaScript grows from 693 lines in the two previous modules to 861 lines across repeater-query.js, query-fields.js, and repeater-sources.js. The PR description should also be corrected: master...HEAD is +2,093/-1,436, or +657 lines overall, rather than “Net -400 lines.” I think the async lifecycle and mode-filter inconsistencies should be addressed before merge, then the team can decide whether future source variability justifies the descriptor approach.

web/js/ui/repeater-query.js
jasiek: These are very fast, nothing to worry about here.

web/js/ui/repeater-query.js
jasiek: Not a big deal.

web/js/ui/repeater-sources.js
jasiek: Fair point - if nothing is selected, then by default, query only FM repeaters. Add a test for this.

jasiek: Fixed in 40ff651: an empty mode selection now falls back to analogue only instead of "any mode", so dmr/p25/nxdn/m17 records stay out even on radios that advertise those modes. Covered by a new test ("an RSGB query with no mode selected falls back to analogue only") with a DMR-capable harness radio.

PR #71, opened 2026-08-27
Add IRTS repeater directory queries

jasiek:
## Summary

- add an IRTS repeater query action backed by api.codeplug.org/irts
- load country, band, and mode filters from the IRTS metadata endpoint
- preserve the shared RXF import flow while handling IRTS radio-perspective RX/TX frequencies and DMR/Fusion mode labels
- cover endpoint configuration, modal behavior, disabled deployments, and representative EI2TRR channel construction

## Validation

- npm test
- npm run check:python
- live api.codeplug.org/irts query and production-origin CORS check

## Dependencies

No new dependencies.


web/js/ui/repeater-query.js
```diff
@@ -44,10 +44,10 @@ const FIELD_FACTORIES = {
...
 
-  // Online repeater queries against przemienniki.net/RepeaterBook depend on a
-  // CORS proxy; when none is configured (blank base) the endpoints are null
-  // and those sources are disabled: their menu items are hidden so they can't
-  // fire requests that will fail.
+  // Online repeater queries against przemienniki.net, RepeaterBook and IRTS
+  // depend on the configured API base; when none is configured (blank base)
+  // the endpoints are null and those sources are disabled: their menu items
+  // are hidden so they can't fire requests that will fail.
   const endpoints = buildRepeaterEndpoints(resolveRepeaterApiBase());
```
jasiek, line 51: This endpoint should always be available, so it's okay to assume that it will always be there, no need to be super cautious, we can let the request fail, it gets us an additional testing scenario.

web/js/ui/repeater-sources.js
```diff
@@ -59,10 +60,18 @@ export function createRepeaterSources(ctx, { endpoints }) {
...
   // query-parameter API, same XML response format.
-  function remoteDirectorySource({ key, label, actionLabel, insertLabel, menuButton, sourceEndpoints }) {
+  function remoteDirectorySource({
+    key,
+    label,
+    actionLabel,
+    insertLabel,
+    menuButton,
+    sourceEndpoints,
+    qrgPerspective = "repeater",
```
jasiek: The perspective parameter can be retrieved from the list of repeaters under //rxf/perspective - if it is missing, signal an error to the user.

web/js/datasources.js
```diff
@@ -325,9 +341,11 @@ export function buildPrzemiennikiRows(repeaters, { createBlankRow, setRowValue,
       DSTAR: ["DV", "DSTAR", "D-STAR"],
       ATV: ["ATV"],
       ECHOLINK: ["ECHOLINK", "FM", "NFM", "FMN"],
+      DMR: ["DMR", "MOTOTRBO", "DIG"],
       MOTOTRBO: ["DMR", "MOTOTRBO", "DIG"],
       APCO25: ["P25", "APCO25", "APCO-25", "DIG"],
       C4FM: ["C4FM", "DN", "VW", "DIG"],
+      FUSION: ["DN", "C4FM", "VW", "DIG"],
```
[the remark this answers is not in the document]
jasiek: Fix this.

web/js/ui/repeater-sources.js
jasiek: Addressed in 1a1fdb8: the RXF parser now reads rxf > perspective, rejects missing or unknown values, and passes the declared perspective into frequency normalization.

web/js/datasources.js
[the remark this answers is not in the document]
jasiek: Addressed in 1a1fdb8: protocol-specific digital mappings no longer fall back to generic DIG or the blank row default mode. Unsupported rows are skipped, logged, reported in status, and covered by tests.

web/js/ui/repeater-query.js
jasiek: Addressed in 1a1fdb8: a blank API-base override now hides only the proxy-dependent sources. IRTS remains available through https://api.codeplug.org/irts, and request failures are surfaced to the user.

PR #117, opened 2026-09-05
Keep each channel in its own memory slot when editing the grid

### LINKED ISSUE #73: Editing the grid renumbers Location from 0, so upload fails on 1-based radios and re-slots channels on 0-based ones
## Summary

`reindexLocationColumn()` rewrites every row's `Location` to its 0-based array index after Insert, Remove, Move Up/Down, Cut, Paste and CSV merge-import. The upload path requires each `Location` to fall inside the driver's `memory_bounds`, and 117 drivers are 1-based (UV-5R is `(1, 128)`). One edit therefore makes Upload fail. On 0-based radios the same code compacts gaps, so channels stored at memories 1, 5 and 50 are rewritten into slots 0, 1 and 2 and memory 50 is erased.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/channel-table.js#L155-L162 (`row.Location = String(idx)`)
- Callers: `channel-table.js` lines 234, 260, 285, 330, 481 and `web/js/ui/codeplug-io.js:96`
- Upload check: https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L1037-L1045
- `validate_rows_for_upload` never checks bounds, so preflight passes and the failure only appears mid-transfer.

Yes. Cited lines re-read and cross-checked against `chirp/chirp/drivers/uv5r.py:2302` (`memory_bounds = (1, 128)`). Not reproduced at runtime.

## Reproduction

1. Select Baofeng UV-5R, connect, Download.
2. Click Insert once (or Remove, Move, Cut/Paste).
3. Click Upload.

Expected: upload succeeds. Actual: "Channel Location 0 is outside radio memory bounds" after preflight passed.

For a 0-based radio: download a codeplug with channels at non-contiguous memories, Insert once, note every `Location` is now contiguous from 0, so an Upload writes channels into different slots than they came from.

## Priority

Critical: affects every radio after any grid edit, and on 0-based radios silently writes channels to the wrong memories.

jasiek:
Fixes #73.

## The problem

`reindexLocationColumn()` rewrote every row's `Location` to its 0-based array index after Insert, Remove, Move Up/Down, Cut, Paste and CSV merge-import. But a `Location` is the memory slot a channel occupies, not its position in the list, and two things followed from conflating them:

- **Upload failed on 1-based radios.** 147 of the driver `memory_bounds = (` call sites number memories from 1 (against 65 from 0), so allocating from 0 produced a Location the radio has no memory for. `_apply_rows_to_radio_instance` rejects it — but only partway through the clone, with the radio already open and some memories written.
- **Sparse codeplugs were silently compacted on 0-based radios.** Codeplugs read from a radio are routinely sparse: `chirp/tests/images/Baofeng_UV-5R.img` fills 37 of its 128 slots, at 0-1, 25-31, 50-66, 80-86 and 124-127. Loading it and clicking Insert once collapsed all 37 onto 0-36, moving HTAC1 off memory 25 and VCALL from 124 to 33. The upload then wrote them to those slots without complaining, which matters because scan lists and the channel dial address memories by number.

Verified at the driver level against the real image — replaying the old renumber-then-write produces `[0..36]` where the original occupancy was `[0, 1, 25, 26, ..., 124, 125, 126, 127]`.

Note the numbering shown when that image loads (0, 1, 25, 26, …) was never the bug — that is the radio's own layout, faithfully preserved. The damage only started once the grid was edited.

## The fix

Allocation goes through `assignFreeLocations()`, which leaves every row holding a valid unclaimed Location where it is and gives the rest — blank inserts, pasted rows past the end, imported rows colliding with what is loaded — the lowest free slot inside the driver's bounds. Bounds come from the Location column's `min`/`max`, which `get_radio_column_metadata()` already fills from `memory_bounds`; with no radio selected, `0..` applies.

Per operation:

- **Insert / bulk insert** (repeater queries, band-plan presets): the new rows take free slots, everything else stays put.
- **Remove**: reassigns nothing — removing a channel frees its memory and moves no other channel.
- **Move Up/Down**: the slots stay where they are and the channels rotate through them, so a move swaps two channels' memories rather than renumbering. The occupied set is unchanged, so a sparse layout stays sparse.
- **Paste-overwrite**: the pasted channel inherits the slot it lands on, not the Location it was copied from — pasting between codeplugs must not drag the source's numbering across.
- **CSV merge-import**: imported rows keep the Location the file gave them unless it is out of bounds or already taken.

`validate_rows_for_upload` additionally checks Location range and uniqueness, so a bad value surviving from an imported CSV is a highlighted cell at preflight rather than a failure mid-transfer. A non-integer Location already raised from `_memory_from_row_values`, so that case is left to the existing error.

## Tests

New `scripts/test-ui-channel-locations.mjs` (7 cases, registered in `test:channels`), driving the real UI through the existing fake-DOM harness on a codeplug shaped like the UV-5R image:

- a sparse codeplug loads with the radio's own numbering
- insert takes the lowest free memory and leaves every other channel in place
- insert on a 1-based radio never allocates memory 0
- removing a channel frees its memory and moves no other channel
- move swaps two channels' memories instead of renumbering
- paste-overwrite gives pasted channels the memory they overwrite (and the confirmation names the real slots)
- paste past the end allocates free memories

Six of the seven fail on the pre-fix code and pass after; the load test passes on both, since that path was already correct.

`scripts/test-channel-list.mjs` gains a preflight case covering out-of-bounds and duplicate Locations.

Full suite green: 959 tests, 0 failures. `npm run check:python` clean.

## Dependencies

None added.


jasiek:
Pushed a follow-up: **the grid is now kept in memory order at all times.**

Splitting Location from the row index created a gap this PR had not closed. Inserting into a codeplug whose memories run 0-10 and resume at 30 correctly took memory 11, but the row stayed where it had been spliced — below memory 31 — so the list stopped reading as the radio's memory map. That ordering was previously free, because Location *was* the row index; it now has to be maintained.

`reconcileLocations()` is the single call every row operation and every load makes after touching `state.currentRows`: `assignFreeLocations()` (unchanged) hands out slots, then `sortRowsByLocation()` restores the ordering. Rows with no usable Location sort last, keeping their arrival order, and stay the preflight's problem to report. Splice position no longer decides anything for inserts, so they push and sort; operations reselect by row identity rather than by the index they spliced at, since the list is reordered underneath them. Loads reconcile too — a radio or image always yields ascending memories, but a hand-written CSV need not.

Two new tests, plus two updated expectations that were pinning the old placement:

- an inserted channel sorts into its memory rather than onto the end (the 0-10 / 30-31 case above)
- a codeplug listing channels out of order loads in memory order

Suite: 961 tests, 0 failures.

web/python/runtime_bridge.py
```diff
@@ -764,15 +764,65 @@ def _infer_csv_error_column(error_text: str):
...
+    if not bounds:
+        return None
+    try:
+        lo, hi = bounds
+        return int(lo), int(hi)
+    except Exception:
+        return None
+
+
 def validate_rows_for_upload(rows, module_name="", class_name=""):
```
jasiek: I need to understand what each row looks like, add type definition for rows which is an array (?) of row objects.

jasiek: Add type annotations wherever else we refer to a collection of rows, or a row.
Add a type checking step for this file.

jasiek: Done in 2dc9748. `Row = dict[str, Any]` / `Rows = list[Row]` are defined next to `CSV_HEADERS` with the shape spelled out: one JSON object per channel, keyed by CSV header name, values are text — `"Location": "25"`, `"Frequency": "443.000000"`. Rows cross the boundary as `JSON.stringify` in `setRowsJsonGlobal()` and `json.loads` here, so the keys are exactly the CSV headers.

The value type is `Any` rather than `str` deliberately, and the comment says why: a row can also carry non-header keys the editor rides along on it — currently the `__geo` sidecar from `web/js/row-geo.js`, which is an object. Every consumer in this file projects a row through `CSV_HEADERS` and ignores everything else, which is what keeps that sidecar out of a codeplug.

jasiek: Done in 2dc9748.

**Annotations**: every function taking or returning rows — `_memories_from_rows`, `normalize_rows`, `validate_rows_for_upload`, `_radio_rows_from_instance`, `_apply_rows_to_radio_instance`, `_upload_selected_radio_sync`, `upload_selected_radio`, `export_image_base64` — plus the row-adjacent ones: `parse_csv`, `_row_values_for_csv`, `_memory_bounds_for_driver`, `_infer_csv_error_column`, `download_selected_radio`, `load_image_base64`. Row-shaped locals are annotated where rows are built, and the preflight now has a named `ValidationIssue` type for its issue list.

**Type checking**: `npm run check:types` runs pyright over `web/python`, wired into `npm test`. It is an npm devDependency (pinned 1.1.413) rather than a pip install, so CI picks it up from the existing `npm ci` with no Python tooling step.

Two things pyright cannot see across, handled by declaring rather than blanket-suppressing where that was possible:

- `web/python/typings/js.pyi` declares the JS serial bridge (`serial_open`, `serial_write_bytes`, `serial_progress`, …). Worth having beyond the checker: it is now the one place the JS and Python halves of that contract are written down together.
- `reportAttributeAccessIssue` is off. This module deliberately builds fake modules — the gettext builtins and the pyserial shim assign attributes onto `ModuleType` — and imports `chirp`, which only exists inside Pyodide. I preferred this to ~19 inline ignores or to rewriting working shim code to satisfy a checker. Argument, return, assignment and optional checks all stay on.

I confirmed the step is not vacuous: passing a `str` where `Rows` is expected, and returning one from `_radio_rows_from_instance`, both fail the check.

jasiek:
Review feedback actioned in 2dc9748 (replies on both threads), then I went looking for what the ordering change broke. It broke three things. All three are fixed in 61c254c, each pinned by a test that fails without it.

**1. Loading silently rewrote the Locations a file gave.** The worst of the three. Sorting on load dragged slot *assignment* along with it, so a CSV listing memories 5, 5 and 9000 loaded as:

```
file said : 5, 5, 9000
user got  : 0, 1, 5     (and reordered)
```

That edits the user's data on the way in, and hides the exact mistake the new preflight checks exist to report. It also made the duplicate-Location check unreachable from the UI. Loads now call `sortRowsByLocation()` alone; only edits reconcile. A file's Locations are the user's data — a duplicate or out-of-bounds one survives to the preflight and gets named there.

**2. A pasted block scattered across the radio.** Paste walked consecutive *rows*, which on a sparse codeplug are not consecutive memories. Three rows pasted at memory 26 landed on **26, 124 and 2**, silently overwriting VCALL at 124 — a channel the confirmation named but the user never aimed at. Paste now targets consecutive memories from the selected channel's own memory:

```
before: 0, 1, 25, [P1@26], [P3@2 !], 124 -> P2 overwrote VCALL
after : 0, 1, 25, P1@26, P2@27, P3@28, VCALL@124
```

Only memory 26 was occupied, so that is the only channel the confirmation names and the only one overwritten. With nothing selected there is no anchor and the block still goes to the lowest free memories, as before.

**3. Move lost its shift-click anchor.** Reselecting by row identity always anchored on the first row; it now stays on the edge the selection travelled towards, as it did before. Caught only after I noticed my first test for it passed against the broken code — a single moved row cannot tell the two anchors apart.

Suite: 964 tests, 0 failures, pyright clean.

**One behaviour change that is a consequence, not a bug, but worth naming:** Insert no longer honours the selection. It always takes the lowest free memory. Previously selecting a row and inserting put the new channel *there* — but only because Location was the row index, which is the thing this PR removes. If placing a new channel near a selected one matters, the natural form is "first free memory at or after the selected channel's memory", and I'd rather add that deliberately than have it fall out of an implementation detail. Say the word and I'll do it.

Issue #91, opened 2026-09-05
Re-selecting a radio after loading an .img keeps the image's driver schema and settings (stale lastLoadedRadioKey)


jasiek:
## Summary

`state.lastLoadedRadioKey` is only written inside `reloadForSelectedRadio()`. Loading an `.img` selects a radio through `selectRadioByDetectedImage()` and `loadSelectedRadioMetadata()` without touching it. So: pick Baofeng UV-5R from the dropdown (key recorded), drop a Yaesu FT-60 image (selected radio, metadata and settings all become FT-60), then re-select UV-5R in the dropdown. `reloadForSelectedRadio()` sees `key === lastLoadedRadioKey` and returns early, so the grid keeps FT-60 column options and bounds and the Settings tab keeps FT-60's groups. A later Upload sends FT-60 settings groups to the UV-5R driver and is blocked with "Setting is not available for this radio image".

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/radio-catalog.js#L228-L241
- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/codeplug-io.js#L430-L436

## Verified against source code

Yes. Only one write site for `lastLoadedRadioKey` (`grep -n lastLoadedRadioKey web/js/ui/*.js`). Not reproduced at runtime.

## Reproduction

1. Select Baofeng UV-5R in the Make/Model dropdowns.
2. Drag a Yaesu FT-60 `.img` (from `chirp/tests/images/`) onto the page.
3. Re-select Baofeng UV-5R in the model dropdown.

Expected: UV-5R columns and settings. Actual: FT-60 column options, bounds and settings remain.

## Suggested fix

Set `lastLoadedRadioKey` wherever metadata is loaded (in `loadSelectedRadioMetadata`), or clear it when an image selects a radio.

## Priority

High: leaves the editor validating against the wrong radio.

PR #122, opened 2026-09-05
Fix stale radio metadata and settings race

### LINKED ISSUE #111: Switching A → B → A while B loads lets B overwrite A's schema and settings
## Summary

reloadForSelectedRadio() returns early when the selection equals lastLoadedRadioKey, before advancing the request token. If A was fully loaded, B starts loading, and the user returns to A before B completes, B's pending response remains valid. It can apply B's schema/settings while the dropdown says A; the completion handler then records A as loaded.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/radio-catalog.js#L223-L245

Yes. Reproduced through the repository's UI harness with deferred runtime responses: load Fast fully, start Slow, reselect Fast, resolve Slow. Selected radio remained Fast while rendered headers became Slow's. No physical radio is needed.

## Reproduction

1. Select a radio A through the make/model dropdown and wait for loading to finish. If A was the initial restored radio, first select another radio, then A, to exercise the completed selection-load path.
2. Enable network throttling in browser DevTools.
3. Select a different radio B whose driver has not yet been loaded.
4. Before B finishes loading, select A again.
5. Wait for B's pending request to finish and inspect the grid's columns/options and Settings tab.

Expected: A retains its own metadata/settings and pending B responses are ignored.
Actual: B can overwrite the editor under A's selection. Timing is required; the deterministic harness reproduction confirms the race.

## Suggested fix

Invalidate pending work on every selection transition and coordinate metadata/settings application using captured radio identity. Restore A's complete state if B has partially applied before reselection; incrementing a token alone does not restore partial changes.

## Related issues

Related to #91, but independently reproducible without loading any image. #91 concerns lastLoadedRadioKey not being updated by image loading; updating that image path does not invalidate B's pending request in this race.

## Priority

High: leaves the editor using another radio's schema/settings.

jasiek:
## Summary

- invalidate pending radio loads on every selection transition, including a return to the last fully loaded radio
- fetch metadata and settings for a captured radio identity and apply them atomically only while that load remains current
- add deterministic A → B → A coverage for both metadata-first and settings-first completion orders
- preserve image-backed settings when reselecting the current radio

Closes #111.

## Regression assessment

An independent review found no implementation regressions in initialization, stale-error handling, settings preservation, or the issue #91 image-import path. The new test explicitly proves the final return to A takes the no-request short circuit, then confirms neither pending half of B can overwrite A.

A pre-existing behavior remains: while a newly selected radio is loading, the editor retains the previous radio state and actions are not gated. This change keeps that state internally consistent instead of allowing a mixed schema/settings view.

## Testing

- npm run check:syntax
- npm run check:types
- npm run test:channels (280 passed)
- npm run test:webusb (121 passed)
- npm run test:settings (555 passed)
- npm run test:build (11 passed)
- focused radio-loading and image-import UI tests (18 passed)
- git diff --check

## Dependencies

No new dependencies.


web/js/ui/settings-panel.js
```diff
@@ -375,22 +375,16 @@ export function createSettingsPanel({ dom, state, log, actions }) {
     }
   }
 
-  async function load(options = {}) {
-    const loadToken = options.loadToken ?? nextRadioLoadToken(state);
-    if (!state.selectedRadio) {
-      settingsState = {
+  async function fetchForRadio(radio) {
```
jasiek, line 378: Functions need comments around what they do and why.

web/js/ui/radio-catalog.js
```diff
@@ -367,24 +378,34 @@ export function createRadioCatalog(ctx) {
...
+      className: radio.className,
     });
-    if (isStaleRadioLoad(state, loadToken)) {
-      return;
-    }
-    state.radioMetadata = meta || { headers: [], columns: {} };
+    return metadata || { headers: [], columns: {} };
+  }
+
+  function applyRadioMetadata(metadata) {
```
jasiek, line 392: Newly added functions need comments around what they do and why

PR #123, opened 2026-09-05
Validate channel writes with CHIRP driver policy

### LINKED ISSUE #83: Upload path skips validate_memory, filter_name and immutable-field checks before set_memory
## Summary

Rows are only checked against global `chirp_common` tables (`_memory_from_row_values`, `validate_rows_for_upload`), never against the driver's `RadioFeatures` (`valid_bands`, `valid_tones`, `valid_tuning_steps`, `valid_characters`, `valid_duplexes`, `valid_cross_modes`) or `mem.immutable`. CHIRP runs `validate_memory`, `filter_name` and `check_set_memory_immutable_policy` before every write (`chirp/chirp/wxui/memedit.py:1978, 2045`; `import_logic.py` around line 290). Drivers assume validated input.

Concrete case: `uv5r.py:1058-1062` marks GT-5R out-of-band channels `immutable=["duplex", "offset"]`; the bridge lets the user set duplex/offset and `set_memory` (`uv5r.py:1097-1109`) writes the TX frequency unconditionally, re-enabling transmit where the type-accepted driver forbids it. Names with characters outside `valid_characters` and tones outside `valid_tones` are written raw.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L1023-L1058

Yes by static review (medium-high confidence); the GT-5R immutable case cross-checked in `uv5r.py`. Not reproduced at runtime.

## Reproduction

1. Load a Radioddity GT-5R image with a channel outside the GMRS band.
2. Set Duplex to "+" and Offset to 5.000000 on that row (the grid allows it).
3. Upload or Export Binary and reload.

Expected: the edit is rejected with the driver's immutable-field message. Actual: the driver writes the TX frequency.

## Suggested fix

Run `radio.validate_memory(mem)` (fail on errors, surface warnings), `radio.filter_name` and the immutable-policy check per row in `_apply_rows_to_radio_instance`, and surface the results through `validate_rows_for_upload` so preflight catches them.

## Priority

High: can produce codeplugs the driver would have refused.

jasiek:
## Summary

- run CHIRP name filtering, immutable-field policy, and selected-driver validation before upload and binary export
- surface blocking validation errors in the grid and non-blocking driver warnings in Debug Output
- preserve unchanged legacy memories and serialize cached images through CHIRP's file path

## Regression investigation

- exercised the validation/apply path against all 358 loadable images in the 359-image upstream corpus; no apply failures remained (the existing Kenwood NX-800 fixture still fails earlier in settings loading)
- compared the changed path against the previous behavior across resolvable images; found no current-fail/baseline-success cases after fixes
- added coverage for migrated Icom images and Retevis radios whose immutable power values are lossy in the grid

## Validation

- npm test (all syntax, type, channel, WebUSB, and settings tests passed; build tests were rerun outside the filesystem sandbox because they bind localhost)
- npm run test:build
- Python bytecode compilation for web/python/runtime_bridge.py
- git diff --check

## Dependencies

No new dependencies.

Closes #83

### DIFF
--- a/FINDINGS.md
+++ b/FINDINGS.md
@@ -52,6 +52,7 @@ CHIRP `80d93fe5`.
 
 *Pinned by tests:*
 - Our CSV export is byte-identical to CHIRP's own across the image corpus, an import invents no channel 0, and a DV row keeps the documented column layout; the band-keyed power divergences are tracked as an explicit list that fails when one appears or disappears — `test-csv-export-parity.mjs`.
+- CHIRP's GT-5R driver marks out-of-band duplex/offset fields immutable but clears that list inside its bulk-import policy hook; matching the desktop grid requires enforcing the declared fields before calling the hook — `test-channel-list.mjs`.
 - Driver validation only runs for changed rows, because legacy codeplugs can contain unchanged memories their current driver rejects. Cached images must use CHIRP's file load/save path rather than `get_mmap()`: high-bit-flipped Icom drivers return a wire-format map there, which is not valid file input — `test-channel-list.mjs`, `test-channel-round-trip.mjs`.
 - `really_from_csv()` cannot read back what `to_csv()` writes (it ignores the Power column and rejects `split`/`off` duplex), so `_memory_from_row_values()` parses rows directly: power resolves to the driver's *own* `PowerLevel` object, frequencies parse with `parse_freq()`, and an unset power stays unset — `test-channel-round-trip.mjs`, whose header carries both original defects in full.
 - Power labels publish the wattage CHIRP would export, duplicates included, and the grid spells the table out as a hover legend — `test-power-level-watts.mjs`.
--- a/FINDINGS.md
+++ b/FINDINGS.md
@@ -52,6 +52,7 @@ CHIRP `80d93fe5`.
 
 *Pinned by tests:*
 - Our CSV export is byte-identical to CHIRP's own across the image corpus, an import invents no channel 0, and a DV row keeps the documented column layout; the band-keyed power divergences are tracked as an explicit list that fails when one appears or disappears — `test-csv-export-parity.mjs`.
+- Driver validation only runs for changed rows, because legacy codeplugs can contain unchanged memories their current driver rejects. Cached images must use CHIRP's file load/save path rather than `get_mmap()`: high-bit-flipped Icom drivers return a wire-format map there, which is not valid file input — `test-channel-list.mjs`, `test-channel-round-trip.mjs`.
 - `really_from_csv()` cannot read back what `to_csv()` writes (it ignores the Power column and rejects `split`/`off` duplex), so `_memory_from_row_values()` parses rows directly: power resolves to the driver's *own* `PowerLevel` object, frequencies parse with `parse_freq()`, and an unset power stays unset — `test-channel-round-trip.mjs`, whose header carries both original defects in full.
 - Power labels publish the wattage CHIRP would export, duplicates included, and the grid spells the table out as a hover legend — `test-power-level-watts.mjs`.
 - Values pasted into columns the radio marks read-only keep their value, and numeric enum columns match spreadsheet-normalized values (`5` for `5.00`, `23` for `023`) — `test-ui-channel-cut.mjs`, `test-clipboard.mjs`.
--- a/web/python/runtime_bridge.py
+++ b/web/python/runtime_bridge.py
@@ -11,7 +11,7 @@ import re
 import sys
 import tempfile
 import types
-from typing import Any, Optional, Sequence
+from typing import Any, Literal, Optional, Sequence
 
 sys.path.insert(0, "/webchirp_runtime")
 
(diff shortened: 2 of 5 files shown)


web/python/runtime_bridge.py
```diff
@@ -802,12 +803,179 @@ def _memory_bounds_for_driver(
...
 
 
+def _radio_instance_for_row_validation(module_name: str, class_name: str):
+    """Build the same image-backed radio that a later upload/export will use."""
+    return _best_effort_radio_instance(
+        module_name, class_name, require_cached=False
+    )
+
+
+def _immutable_field_errors(existing, new) -> list[Exception]:
```
jasiek: Add type annotations.

web/python/runtime_bridge.py
```diff
@@ -802,12 +803,179 @@ def _memory_bounds_for_driver(
...
+        if getattr(existing, field) != getattr(new, field):
+            immutable_errors.append(
+                chirp_common.ImmutableValueError(
+                    f"Field {field} is not mutable on this memory"
+                )
+            )
+    return immutable_errors
+
+
+def _preserve_unedited_immutable_fields(row: Row, existing, mem) -> None:
```
jasiek: Add type annotations to all added code.

web/python/runtime_bridge.py
```diff
@@ -802,12 +803,179 @@ def _memory_bounds_for_driver(
...
+def _row_matches_memory(row: Row, memory) -> bool:
+    """Compare a grid row at exactly the fidelity exposed by the grid."""
+    row_values = [str(row.get(header, "") or "") for header in CSV_HEADERS]
+    memory_values = [str(value) for value in _row_values_for_csv(memory)]
+    return row_values == memory_values
+
+
+def _validation_column(message: Any) -> str:
+    """Map a CHIRP validation or immutable-field message to a grid column."""
+    text = str(message or "")
```
jasiek, line 1000: What can a message be here? str only or None? Annotate accordingly.

web/python/runtime_bridge.py
```diff
@@ -861,7 +1030,52 @@ def validate_rows_for_upload(
                     "message": error_text,
                 }
             )
-    return {"valid": len(issues) == 0, "issues": issues}
+            continue
+
+        if (
```
jasiek, line 1094: We need comments for this section to indicate why we need to do all this. Better yet, roll this into a function.

web/python/runtime_bridge.py
```diff
@@ -1124,7 +1380,19 @@ def _apply_rows_to_radio_instance(
             )
         seen_numbers.add(number)
         freq_text = str(row.get("Frequency", "") or "").strip()
+        existing = radio.get_memory(number)
+        if _row_matches_memory(row, existing):
+            continue
         if not freq_text:
+            if existing.empty:
```
jasiek: Need comments around why we need to do this.

FINDINGS.md
```diff
@@ -52,6 +52,7 @@ CHIRP `80d93fe5`.
 
 *Pinned by tests:*
 - Our CSV export is byte-identical to CHIRP's own across the image corpus, an import invents no channel 0, and a DV row keeps the documented column layout; the band-keyed power divergences are tracked as an explicit list that fails when one appears or disappears — `test-csv-export-parity.mjs`.
+- CHIRP's GT-5R driver marks out-of-band duplex/offset fields immutable but clears that list inside its bulk-import policy hook; matching the desktop grid requires enforcing the declared fields before calling the hook. Driver validation only runs for changed rows, because legacy codeplugs can contain unchanged memories their current driver rejects. Cached images must use CHIRP's file load/save path rather than `get_mmap()`: high-bit-flipped Icom drivers return a wire-format map there, which is not valid file input — `test-channel-list.mjs`, `test-channel-round-trip.mjs`.
```
jasiek: Split this PR into two commits, one containing logic related to the GT-5R and the other for everything else.

web/python/runtime_bridge.py
jasiek: Added an explicit chirp_common.Radio return annotation here.

FINDINGS.md
jasiek: Rewrote the branch into two commits: 2db11af contains general CHIRP validation, filtering, warnings, typed helpers, and image handling; b2ece04 contains declared-immutable enforcement and the GT-5R regression test.

web/python/runtime_bridge.py
jasiek: Added annotations to every new helper, including radio and Memory parameters, image bytes, action literals, validation messages, and tuple return types. Pyright passes with zero errors.

web/python/runtime_bridge.py
jasiek: Extracted the duplicated logic into the typed _prepare_row_change helper. Its docstring explains current-memory comparison, untouched legacy-row skipping, and why an empty frequency is validated as an erase.

web/python/runtime_bridge.py
jasiek: The apply path now delegates this decision to _prepare_row_change, and the remaining call-site comment explains why the current driver memory must be loaded before applying a row. Omitted-row erasure has a separate policy rationale comment.

web/python/runtime_bridge.py
jasiek: A validation message can be a CHIRP ValidationMessage, which is a str subclass, or an Exception caught while invoking driver policy or validation. I added the ValidationMessage = str | Exception alias and annotated this parameter with it.

PR #126, opened 2026-09-05
Fix TH-D72 clone abort by honouring driver DTR/RTS changes

### LINKED ISSUE #77: WebSerialPipe.setRTS() requires an argument, so Kenwood TH-D72 clones die with TypeError; DTR/RTS setters never reach the port
## Summary

The pyserial shim declares `def setRTS(self, value)`. `thd72.py:480` calls `self.pipe.setRTS()` with no argument and only catches `AttributeError`, so the call raises `TypeError: missing 1 required positional argument` and the download or upload aborts before the first block. Independently, `setRTS`/`setDTR` and the `rts`/`dtr` attributes only store a boolean; nothing propagates them to the port, so mid-clone control-line toggles (`thd72.py`, `ft450d.py`) are inert. Only `_prepare_clone_session` touches the lines, once, before the clone.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L864-L870
- Driver: `chirp/chirp/drivers/thd72.py:478-484` and `:521-523`

Yes. Shim and driver re-read. Not reproduced on hardware.

## Reproduction

Requires a Kenwood TH-D72: select it, connect, Download. Expected: clone starts. Actual: `TypeError` in the debug panel before any data transfers.

## Suggested fix

Give `value` a default of `True` (pyserial semantics) and forward `setRTS`/`setDTR` and the `rts`/`dtr` properties to the JS bridge's `setSignals`.

## Priority

High for the TH-D72 (hard crash); medium for the inert line toggles.

jasiek:
Fixes #77.

## What was broken

The pyserial shim declared setRTS(self, value). thd72.py:480 calls self.pipe.setRTS() with no argument and catches only AttributeError, so the download or upload died with a TypeError before the first block. Separately, setRTS/setDTR and the rts/dtr attributes only stored a boolean, so every mid-clone control-line toggle was inert: only _prepare_clone_session touched the port, once, before the clone.

## What changed

- **web/python/runtime_bridge.py** — setRTS/setDTR take pyserial's optional level (default True), and rts/dtr become writable properties. Both spellings now push the current line state to the transport via a new serial_set_signals bridge call. Failures are logged to the debug panel rather than raised, since control lines are advisory and several adapters cannot change them.
- Pipe construction seeds the line state *without* touching the port — _prepare_clone_session() still owns the initial assertion together with its settle delay. The four lines repeated at three clone entry points are now one _new_serial_pipe() helper, so the pipe a driver sees is built identically everywhere.
- **web/js/serial.js** — BrowserSerialBridge.setSignals() plus a setSignals RPC op. A null line is left untouched, so changing RTS alone does not clear DTR.
- **web/js/runtime-rpc.js** — installs the serial_set_signals global next to the other bridge globals.
- **web/python/typings/js.pyi** — declares the new bridge function, keeping both halves of the contract written down together.

## Tests

- New scripts/test-serial-control-lines.mjs (registered in test:webusb): line selection, refusal without an open port, RPC forwarding, and that a port which cannot set signals logs and continues instead of failing the clone.
- scripts/test-pyserial-shim.mjs gains the Python half — a bare setRTS() defaults to asserted, property assignment reaches the transport, and the constructor's seeding does not. The stub bridge in the test harness records every setSignals call so this is observable.
- Full suite green: check:syntax, pyright, 981 tests across test:channels / test:webusb / test:settings / test:build.

## Known remaining gap for the TH-D72

Not fixed here, and recorded in FINDINGS.md: thd72 also does self.pipe.baudrate = 57600 mid-clone, right after 0M PROGRAM. 23 driver call sites assign pipe.baudrate (plus parity/stopbits/bytesize) and the pipe only records the value, because Web Serial has no rate change — reconfiguring means closing and reopening the port, which no bridge op does. So a TH-D72 clone is expected to get further than it did, but not necessarily to complete. That belongs in its own change; happy to open a follow-up issue.

## Dependencies

None added.


jasiek:
Rebased onto master (68b5c93) to reconcile with #125, which landed in the same three files. Resolution:

- **web/python/runtime_bridge.py** — auto-merged cleanly; write() keeps its byte-count return and the pipe constructor keeps the seeded DTR/RTS state.
- **scripts/test-pyserial-shim.mjs** — both sides were appends; kept all four tests.
- **FINDINGS.md** — the two entries were making the same general point from opposite directions, so rather than leave both restating it, mine is renamed to **the-pipe-is-reconfigured-mid-clone**, moved to sit directly beside **shim-owes-pyserial-return-values-not-just-methods**, and now opens by naming it as its sibling (argument side vs. return side). The shared Pinned-by-tests bullet keeps #125 wording with the control-line line added after it.

Full suite green after the rebase: pyright plus 983 tests.

jasiek:
Rebased onto master (d99a1dc) to reconcile with #129, which landed in the same files. Resolution:

- **web/python/runtime_bridge.py** — #129 added `_driver_baud_rate()` (normalises a driver's declared `BAUD_RATE`, which can be missing or non-numeric); this PR added `_new_serial_pipe()` (the one place a clone pipe is built). Composed rather than picking a side: the shared helper now builds the pipe *through* the normaliser, so the three clone entry points get both.
- **package.json** — union of the test lists from both sides.
- **scripts/test-radio-harness.mjs** — kept #129's four-argument `serial_prepare_clone` global alongside this PR's `serial_set_signals`. Worth noting: my first resolution kept *both* `serial_prepare_clone` assignments, which silently shadowed #129's with the three-argument version — `test-clone-baud-plumbing.mjs` caught it. "Keep both" is wrong whenever both sides write the same slot.

Full suite green after the rebase: pyright plus 991 tests.

PR #132, opened 2026-09-05
Let settings read from an image be written straight back

### LINKED ISSUE #34: Settings cannot be re-applied after being read for 13 images
## Summary

Reading a radio's settings and applying them straight back fails for **13 images**. Upload aborts with the generic wrapper:

```
RuntimeUnsupportedError: Radio settings validation failed before upload
```

Behind that wrapper, `_validate_and_apply_radio_settings` (`web/python/runtime_bridge.py:970`) reports four distinct causes. This is the settings-side analogue of #31: in most cases the read path emits values the write path refuses.

### A. Value CHIRP itself emitted is rejected — 1 image

`Retevis_C2.img`:

```
dtmf.dtmf_local_id: Value '1234    ' contains invalid character ` '
```

The trailing-space padding comes from the driver's own stored value. Exactly the `to_csv` / `really_from_csv` asymmetry from #31, in the settings tree.

### B. Current value is not in the setting's own option list — 3 images

```
Retevis_MA1:   basic.backlight   -> "8 is not valid for this setting"
Retevis_MA1:   basic.p1          -> "Enter is not valid for this setting"
Retevis_RA685: advanced.savemode -> "On is not valid for this setting"
Retevis_RA85:  advanced.savemode -> "On is not valid for this setting"
```

The image holds a value the driver does not list as selectable. Reading is fine; writing the same value back is refused.

### C. Setting disappears between read and re-apply — 4 images

```
Icom_IC-F621-2:     group_continuous_tone, group_scan
Quansheng_UV-K5:    basic.language
Retevis_HA2:        aprsinfo.anaptttype, aprsinfo.aprsaudio
Wouxun_KG-UV920P-A: rmt_grp.rc_sw, rmt_grp.rc_power
```

`Setting is not available for this radio image.` — the settings tree built on re-apply does not contain a path that was present when reading. Related to **settings-need-backing-state** in FINDINGS.md: these drivers build the tree conditionally from image content, so a tree serialized from one instance does not necessarily map onto a freshly constructed one.

### D. Validation crashes — 5 images

```
Yaesu_FT-25R, FT-4VR, FT-4XE, FT-65E, FT-65R:
  misc.cw_id  -> object of type 'NoneType' has no len()
  misc.passwd -> object of type 'NoneType' has no len()
```

Not a validation failure but an exception surfacing as one. A setting whose value is `None` reaches a length check. Worth fixing first — it is the only category that is unambiguously a defect on our side rather than a data/driver mismatch.

## Reproduction

```js
const loaded = await harness.loadCodeplugBinary(raw);   // read settings
await harness.runPythonJson(`
_r = _import_radio_class(_m, _c)(memmap.MemoryMapBytes(bytes(
    chirp_common.CloneModeRadio._strip_metadata(base64.b64decode(_b))[0])))
_res = _validate_and_apply_radio_settings(_r, json.loads(_s), apply_changes=True)
json.dumps({"valid": _res["valid"], "issues": _res["issues"]})`,
  { _m: match.module, _c: match.className, _b: b64, _s: JSON.stringify(loaded.settings) })
```

## Suggested order

1. Category D — stop the `NoneType` crash; a `None` value should serialize and validate as empty.
2. Category A — accept values the driver emitted, mirroring the #31 approach of letting the driver's own validation decide.
3. Categories B and C — need per-driver triage; C may not be fixable by replaying a serialized tree at all, and may want re-reading the tree from the same instance instead.

Measured across `chirp/tests/images/` after #32.

jasiek:
Fixes #34.

Reading a radio's settings and applying them straight back failed for **9 of the 231** upstream test images that carry a settings tree, aborting with `RuntimeUnsupportedError: Radio settings validation failed before upload`. After this change: **0 of 231**.

## What the issue got right, and one correction

Categories A (autopadded string) and D (`NoneType has no len()`) are as described. Categories B and C need amending:

- **Category B is already gone.** `Retevis_MA1`, `Retevis_RA685` and `Retevis_RA85` pass on the current submodule pin before any change here, so nothing was needed for them.
- **Category C is not a tree that changes shape.** The trees are *identical* — I diffed every path from the read instance against a freshly built one and they match exactly, in the same order. Instrumenting the swallowed exception behind `Setting is not available for this radio image.` shows two different addressing failures:
  - `kguv920pa.py:770-771` names two sibling groups `rmt_grp` (an upstream typo — the Repeater group should be `rpt_grp`), and `retevis_ha2.py:1212-1213` names two groups `aprsinfo`. `RadioSettings.__getitem__` returns the *first* name match, so everything under the second group was unreachable → `KeyError`.
  - `icf520.get_settings()` returns `list(RadioSettingGroup("top", ...))` — a bare Python `list`, not a `RadioSettings`, which indexes only by integer → `TypeError: list indices must be integers or slices, not str` for every top-level group.

So the underlying defect is that the bridge addressed settings by **name path**, when CHIRP guarantees neither unique sibling names nor a name-indexable root.

## Changes, all in `web/python/runtime_bridge.py`

1. **Address nodes by position, not by name.** `_settings_container_children()` normalizes the three container shapes; `_match_serialized_child()` pairs a serialized child with the node at the same position, verifies the name matches, and falls back to a name scan only if the two trees genuinely differ in shape. Clears the Icom, Wouxun and Retevis HA2 failures.
2. **Skip values CHIRP could not initialize.** A driver that builds a value its own validation rejects gets the error logged and swallowed by `RadioSettingGroup.__init__` (`settings.py:80-90`), leaving `_current` at `None`. That serialized as `null` and was replayed as `null`, reaching `len(None)` inside `set_value()`. Desktop CHIRP strips these before applying (`wxui/settingsedit.py:177-190`); this is the same rule. Clears the five Yaesu FT-4/FT-65 failures.
3. **Do not rewrite a value the driver just emitted.** `RadioSettingValueString.set_value()` autopads to maxlength, and `retevis_c2.py:1189` narrows the charset afterwards — so the driver refuses the space-padded string it had itself produced. `_serialized_value_matches()` compares against the live value and writes only what changed. Clears `Retevis_C2`, and structurally covers the category-B class of failure (a list whose stored index is outside its own options) should it reappear.

Stripping the tree before `radio.set_settings()`, which the issue thread flagged as possibly also needed, turns out not to be: no driver hits `int(None)` once the uninitialized values are no longer replayed. Left out rather than added speculatively.

Note that (3) means an upstream read/write asymmetry is now a *display* concern rather than an upload failure — the UI still shows the padded string or the out-of-range option, and it still cannot be re-selected. That is unchanged by this PR and is a separate, smaller problem.

## Tests

New `scripts/test-settings-round-trip.mjs`, registered in `npm run test:settings`. Beyond asserting the round trip for all 9 images, each of the four upstream conditions has its own fixture assertion — an uninitialized value exists, the trailing pad exists, duplicate group names exist, the bare-list root exists — so a submodule bump that removes one fails loudly instead of leaving the round-trip test passing vacuously.

Each assertion was checked against the defect it claims to catch:
- reverting the bridge fails the round-trip test;
- forcing `_serialized_value_matches()` to always return `True` (i.e. over-skipping) fails the edited-value test, and only that one.

`npm test` passes: 1026 tests, 0 failures. `pyright` reports 0 errors.

## Dependencies

None added.

Model: claude-opus-5


jasiek:
Actioned all five review points in 5417a22 — all were valid, nothing pushed back on.

**Ambiguous name fallback.** Correct, and the risk is real given duplicate names demonstrably exist in this corpus. `_match_serialized_child()` now returns the sole name match or `None`, so an ambiguous name reports a mismatch instead of guessing at a group to write into. Stopping the upload is the safer failure than silently miswriting it.

**Null for an initialized setting.** I measured this before changing it, since the two failure modes point opposite ways: skipping silently keeps the old value, reporting aborts the whole upload — and this PR exists because uploads were aborting on values the user never edited. Across all 231 upstream images with settings, a serialized null and an uninitialized target coincide **in both directions**, zero exceptions (I checked the inverse too). So the null case is unreachable for real driver output, and reporting it can only ever mean a malformed payload. Split the guard: skip on `not target.initialized`, report on a null past it.

**Duplicate-sibling test.** You're right that it tested the wrong thing — pooling top-level ids with children of unrelated parents could pass on a cross-level coincidence, and would keep passing after the duplication it pins was gone. It now walks each sibling array on its own and recurses. Verified non-vacuous: pointed at `Baofeng_UV-5R.img` (no duplicate siblings) it fails.

**Six Pyodide boots.** Fixed; one harness per file. 13.5s → 9.0s.

**Type annotations.** Added on all five helpers — `CLAUDE.md` requires signatures on Python functions, so this was a straight miss.

Re-verified after the changes: all 231 images still round-trip with 0 failures, `npm test` 1026 pass / 0 fail, `pyright` 0 errors.

PR #135, opened 2026-09-06
Choose a radio by searching instead of make/model dropdowns

jasiek:
## What changed

The `Radio make` and `Radio model` dropdowns are gone. Picking a radio is now the search box alone, and a **Selected radio** readout in the sidebar names what the rest of the app will act on.

| | Before | After |
|---|---|---|
| Selection | Two dependent dropdowns + a search box | Search box only |
| Current radio | Implied by the dropdown values | Named in a sidebar readout |
| At startup | First vendor's first model, auto-selected | Nothing selected — the readout says so |
| Live-mode marker | `⚡ AOR AR8200` | `AOR AR8200 ⚡` |
| Clone buttons | Download Radio / Upload Radio | Load from radio / Save to radio |

## Why

**The dropdowns were hiding data the catalog already had.** Every catalog entry carries an `aliases` list built from the driver's CHIRP `ALIASES` — the other vendor/model badges it ships under. A driver is filed under exactly one vendor, so `uv5r:BaofengUV5RGeneric` is also the driver for the Retevis RT5R/RT5RV/RT5, Rugged RH5R, Radioddity UV-5R EX and Ansoko A-5R, and none of those vendors appeared anywhere in the UI: the make list came from `r.vendor` alone. Search now matches those identities and names the one that hit, so a query for `retevis` returns *Baofeng UV-5R — also sold as Retevis RT5R* rather than an unexplained Baofeng.

**Nothing is auto-selected any more.** The old first-vendor default was arbitrary (`iradio_uv_5118` at the current pin), and it quietly loaded a driver nobody asked for on every cold start. The readout now says `No radio model selected` and the serial and clone buttons carry `Search for and select a radio first` until one is picked.

## Reviewer notes

- **The readout does not show the driver class.** It names the radio the way its owner would; `RADIO SELECT Baofeng UV-5R (uv5r.BaofengUV5RGeneric)` still goes to Debug Output on every selection, and the bug-report template still carries it.
- **Element ids are unchanged** — `#radio-download` / `#radio-upload` are internal handles wired through `dom.js`, several test files and the CLI docs, so only the labels moved. The prose naming those buttons (the unsupported-browser cards) follows them.
- **Status-line verbs still say "Download"/"Upload"** (`Upload blocked: 3 invalid value(s)…`). Those name the operation, not the control, and match the analytics event names. Easy to change if you'd rather they follow the buttons.
- **The suggestion list is a real combobox now** (`aria-activedescendant`, per-option `aria-selected`), which matters more once it is the only way in.
- **`analytics` unchanged in shape**: `radio_selected` still reports `method`, now only `search` / `restored` / `image`. The `make` and `model` methods are retired with the dropdowns; no new parameters, so no `CUSTOM_DIMENSIONS` change.
- **The screenshot run had to change.** `npm run screenshots` relied on the boot default to have a radio loaded; it now drives the search box (`scripts/update-screenshots.mjs`, searching `Baofeng UV-5R`). Easy to point at a different radio.
- **Four test files depended on the auto-selection without saying so** — they never picked a radio, they just relied on boot doing it, which is why the Location-bounds and paste tests started failing. Each now selects through the search box, which is closer to what a user does.

## Dependencies

None added.

## Validation

`npm run check:syntax`, `npm run check:types` (pyright, 0 errors) and all of `test:channels` / `test:webusb` / `test:settings` / `test:build` pass — 1031 tests. New coverage pins the alias search and its label, the marker's side, the readout across load/empty/selected, the combobox aria wiring, and that serial actions stay disabled until a radio is selected.

`npm run check:python` could not run in this environment (no `python3` on PATH via asdf); no Python changed.


web/js/ui/radio-catalog.js
```diff
@@ -316,10 +390,8 @@ export function createRadioCatalog(ctx) {
       return false;
     }
     clearRadioFilter();
-    dom.radioMakeEl.value = target.vendor;
-    refreshModelOptions();
-    dom.radioModelEl.value = target.key;
     state.selectedRadio = target;
```
jasiek: [P1] Refresh serial actions after image selection

This path assigns the selected radio but no longer calls actions.updateSerialActionState(). On a fresh session, importing an image with no radio-wide settings therefore leaves Connect, Load, and Save disabled despite the populated readout. The old refreshModelOptions() path performed this refresh indirectly; both the primary and fallback image-selection branches need it.

web/js/ui/radio-catalog.js
```diff
@@ -93,49 +95,143 @@ export function createRadioCatalog(ctx) {
...
+  // implementation detail that belongs in the debug log, which records it on
+  // every selection.
+  function renderSelectedRadio() {
+    const radio = state.selectedRadio;
+    dom.radioSelectionEl.classList.toggle("is-empty", !radio);
+    if (!radio) {
+      dom.radioSelectionNameEl.textContent = catalogStatusText || NO_RADIO_SELECTED_TEXT;
+      return;
+    }
+    dom.radioSelectionNameEl.textContent = radioDisplayLabel(radio);
```
jasiek: [P2] Preserve duplicate-driver identity in the readout

The search suggestion adds className for duplicate vendor/model labels, but the persistent readout always drops it and the search field is cleared after selection. Multiple catalog entries share the same displayed name, so users cannot verify which driver Connect, Load, and Save will act on. Retain the same duplicate-label disambiguator in the selected-radio readout.

jasiek (commented):
I found two regressions in the new selected-radio flow: binary image imports can leave serial controls disabled, and duplicate make/model entries lose their persistent driver disambiguation.

web/js/ui/radio-catalog.js
jasiek: Confirmed and fixed in 0fc7dda.

You were right that `refreshModelOptions()` was doing this indirectly, and that both image branches needed it — so rather than adding two calls, all three selection entry points (suggestion, cookie, image) now go through one `commitSelectedRadio()`:

```js
function commitSelectedRadio(radio) {
  clearRadioFilter();
  state.selectedRadio = radio;
  renderSelectedRadio();
  actions.updateSerialActionState();
}
```

There are now exactly two assignments to `state.selectedRadio` left in the module: the one inside that funnel, and the `= null` in `refreshCatalog()` which already re-gates. So there is no site left that can skip the refresh.

Pinned by a new test in `test-ui-file-drop.mjs`: boot with nothing selected, drop a `.img`, assert Connect goes from disabled/"Search for and select a radio first" to enabled. It fails with the re-gate removed from the funnel — I checked rather than assuming.

web/js/ui/radio-catalog.js
jasiek: Agreed and fixed in 0fc7dda. The readout now appends the class on the same condition the suggestions use, through the same function, so the two cannot disagree about what a radio is called:

```js
function radioLabel(radio, isAmbiguous) {
  const label = radioDisplayLabel(radio);
  return isAmbiguous ? `${label} (${radio.className})` : label;
}
```

One deliberate difference in how the condition is evaluated. The suggestion rows count duplicates within the visible result slice; the readout asks the whole catalog:

```js
function hasAmbiguousLabel(radio) {
  const label = makeModelLabel(radio);
  return state.radioCatalog.some(
    (other) => other.key !== radio.key && makeModelLabel(other) === label,
  );
}
```

A persistent label has no result set to be relative to, so "does anything else in the catalog wear this name" is the right question for it — and it is the stricter of the two, which is the safe direction here.

Pinned by a new test in `test-ui-radio-loading.mjs`: two entries sharing "Acme Twin" read as `Acme Twin (TwinARadio)`, a unique "Acme Only" stays bare. It fails if the readout drops the class unconditionally.

Note the flip side, since it was your earlier ask that the class not be visible: on the ~unique majority of the catalog the readout is unchanged and shows the name alone. The class appears only where the name genuinely cannot identify the driver. Say the word if you would rather see it always.

PR #136, opened 2026-09-06
Report why a CHIRP module could not be fetched instead of ModuleNotFoundError

### LINKED ISSUE #100: ChirpCdnFinder swallows the CDN fetch error and reports ModuleNotFoundError instead
## Summary

`ChirpCdnFinder.find_spec` catches the error from fetching a driver module and returns `None`, so a network failure or 404 surfaces to the user as `ModuleNotFoundError: No module named 'chirp.drivers.x'`. The real cause never reaches the debug panel, against the "preserve debug visibility" rule in CLAUDE.md.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L193-L201

## Reproduction

Block `cdn.jsdelivr.net` after the app has booted, then select a radio whose driver has not been imported yet. Expected: a fetch error naming the URL. Actual: `ModuleNotFoundError`.

## Suggested fix

Log the underlying exception to the debug panel before returning `None`, or re-raise it as an `ImportError` with the URL.

## Priority

Low.

jasiek:
Fixes #100.

## Problem

`ChirpCdnFinder.find_spec` caught every failure from materializing a `chirp.*` module and returned `None`, which hands the import back to `PathFinder`. Its only verdict is `ModuleNotFoundError: No module named 'chirp.drivers.x'`, so three quite different failures arrived wearing the same wrong explanation, and the real cause reached no surface at all:

- a jsDelivr 404 or an offline network (the case in the issue),
- the `RuntimeError: WebAssembly stack switching not supported...` that `_await_js` raises on Safari and pre-152 Firefox — this is the swallowing already recorded in FINDINGS as **no-jspi-browsers-fail-init-as-unavailable**,
- an unimportable driver during the all-driver sweep, where `import_all_driver_modules` could only report *that* a module was skipped, never *why*.

## Change

`find_spec` now logs the cause and its traceback through `_log_debug` (so the debug panel gets it, per the "preserve debug visibility" rule) and re-raises as an `ImportError` naming the module and its source path:

    ImportError: Could not load CHIRP source for chirp.drivers.kguv8d
    from /chirp/drivers/kguv8d.py: Failed to fetch https://cdn.jsdelivr.net/...: 503

Raising the broader `ImportError` rather than the `ModuleNotFoundError` subclass it replaces leaves every caller that catches `ImportError` unaffected; nothing in the runtime or in CHIRP's importable modules matches the subclass specifically.

## Tests

- New `scripts/test-chirp-import-errors.mjs` makes a driver's source fetch reject and asserts the raised error and the debug log both name the underlying failure and carry a traceback, then that a later successful fetch still imports (a failed import must not poison the hook). Verified to fail against the pre-fix finder with exactly the `ModuleNotFoundError` from the issue.
- `scripts/test-metadataless-image-load.mjs` asserted the old `ModuleNotFoundError` text for a nonexistent driver; it now asserts the informative message instead.
- Full suite green: `npm test` — 1056 tests, 0 failures, pyright clean.

## Dependencies

None added.


package.json
```diff
@@ -16,7 +16,7 @@
...
...e --test --experimental-wasm-stack-switching scripts/test-channel-list.mjs scripts/test-ui-radio-loading.mjs scripts/test-call-queue.mjs scripts/test-driver-import-race.mjs scripts/test-chirp-import-errors.mjs scripts/test-clipboard.mjs scripts/test-ui-channel-cut.mjs scripts/test-ui-channel-locations.mjs scripts/test-export-filename.mjs scripts/test-image-metadata.mjs scripts/test-datasources.mjs scripts/test-staticmap.mjs scripts/test-ui-repeater-map.mjs scripts/test-ui-repeater-query.mjs scripts/test-ui-repeater-modal.mjs scripts/test-ui-query-fields.mjs scripts/test-rsgb.mjs scripts/test-ui-file-drop.mjs scripts/test-dom-selectors.mjs scripts/test-debug-log.mjs scripts/test-analytics.mjs scripts/test-analytics-events.mjs scripts/test-sentry.mjs scripts/test-ga-dimensions.mjs scripts/test-gettext-builtins.mjs scripts/test-issue-report.mjs scripts/test-csv-export-parity.mjs scripts/test-power-level-watts.mjs scripts/test-channel-round-trip.mjs scripts/test-pyserial-shim.mjs scripts/test-serial-in-waiting.mjs scripts/test-metadataless-image-load.mjs scripts/test-progress.mjs scripts/test-clone-progress.mjs scripts/test-clone-button-gating.mjs scripts/test-clone-baud-plumbing.mjs",
```
jasiek: [P1] Keep the detect-from-serial regression test when resolving the conflict

This line now conflicts with master: this branch adds scripts/test-chirp-import-errors.mjs, while merged PR #131 added scripts/test-clone-detect-from-serial.mjs to the same test:channels command. Please rebase or merge master and retain both entries; choosing either side wholesale silently removes one regression suite from npm test.

jasiek: Done in 131db7f. Merged origin/master and resolved the conflict by replaying this branch's single insertion onto master's list rather than taking either side, so test:channels now runs both scripts/test-clone-detect-from-serial.mjs (from #131) and scripts/test-chirp-import-errors.mjs. Verified by parsing the merged script and by a full npm test on the merged tree: 342 tests in test:channels, 1075 overall, 0 failures.

PR #137, opened 2026-09-06
Normalise repeater-import CTCSS by the feed's perspective

### LINKED ISSUE #103: CTCSS rx/tx from RXF are not perspective-normalised, unlike the frequencies
## Summary

`buildPrzemiennikiRows` maps `ctcssTx` to `rToneFreq` plus Tone mode and `ctcssRx` to `cToneFreq` regardless of `qrgPerspective`. For IRTS (radio perspective) that is right. For przemienniki.net (repeater perspective, per FINDINGS.md) `type="rx"` is the tone the repeater listens for, which is the one the radio must transmit. A repeater that requires an access tone but transmits none (`<ctcss type="rx">88.5</ctcss>` only) ends up with `cToneFreq` 88.5 and Tone left blank, so the radio never sends the tone. When both directions carry the same tone the swap is invisible, which is why it has gone unnoticed.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/datasources.js#L336-L345
- Frequency perspective handling for contrast: lines 311-316

Yes by static review (medium confidence; depends on przemienniki's ctcss attribute following the same perspective as qrg, which the feed's `perspective` element implies but was not verified against live data). Not reproduced at runtime.

## Reproduction

Query przemienniki.net for an area with a tone-access repeater that does not transmit a tone, import, inspect the row. Expected: Tone "Tone", rToneFreq set. Actual: Tone blank, cToneFreq set.

## Priority

Medium.

jasiek:
Fixes #103

## Problem

RXF's `<perspective>` element labels the whole record, not just `<qrg>`. `buildPrzemiennikiRows` normalised the frequencies by it but read the tones unconditionally: `ctcss type="tx"` to `rToneFreq` plus a Tone mode, `ctcss type="rx"` to `cToneFreq`.

That is correct for IRTS and RepeaterBook (both declare `perspective` = `radio`) and backwards for przemienniki.net (`repeater`), where `rx` is the tone the *repeater* listens for and therefore the tone the *radio* must transmit. A tone-access repeater imported with `cToneFreq` set and `Tone` blank, so the radio sent nothing and never opened the repeater. When both directions carry the same tone the swap has no visible effect, which is why it went unnoticed.

Review then surfaced a second half of the same bug: a tone frequency does nothing without the mode that encodes it. `chirp_common.split_tone_encode` reads `rToneFreq` only under tmode `Tone`, `cToneFreq` only under `TSQL`, and both only under `Cross` — so a receive-only record produced no RX tone at all, and a split pair silently lost its receive half.

## Change

- `buildPrzemiennikiRows` derives `transmitTone`/`receiveTone` from the same perspective flag as the frequencies, so both pairs swap together.
- `applyTonePair` mirrors `chirp_common.split_tone_decode`, CHIRP's own inverse of the same mapping: transmit-only → `Tone`; matching pair → `TSQL`; anything else, receive-only included → `Cross` with `Tone->Tone` or `->Tone`. Each branch writes only the field its mode encodes.
- Cross is gated on the **Tone** column's options, not CrossMode: `valid_cross_modes` stays fully populated even when `has_cross` is false. A radio without Cross keeps the access tone for a split pair, and takes TSQL for a receive-only record.
- `parseCtcssFreq` admits only a positive numeric body. RepeaterBook overloads `<ctcss>` with `CSQ`, `Restricted` and DTCS codes; those were being written through, setting a tone mode backed by the row's default 88.5. DTCS support would need the DtcsCode/RxDtcsCode/DtcsPolarity columns and is out of scope.
- Renamed the option `qrgPerspective` to `perspective` — the old name is what made the tone mapping look out of scope.
- 15 tests in `scripts/test-datasources.mjs` pin both perspectives, every tone shape, both no-Cross fallbacks, and the non-CTCSS bodies.
- FINDINGS.md gains three entries: the perspective covering CTCSS, the non-tone `<ctcss>` bodies, and the tone-needs-its-mode rule.

## Verified against live data

The issue could only infer that `<ctcss>` follows `<qrg>`'s perspective. It does:

- przemienniki.net declares `repeater`; SR4X carries qrg rx 145.000 / tx 145.600 with a lone `<ctcss type="rx">67.0`. 145.000 is the Polish 2m *input*, so `rx` is the repeater's receive side on both elements and that lone tone is an access tone.
- **RepeaterBook declares `radio`, not `repeater`** — only przemienniki.net is repeater-relative among the three.
- Impact: of 646 live Polish records, 77 are access-tone-only (broken before this PR) and 12 tone-transmit-only. Of 1,272 live US 2m RepeaterBook records, 39 carry a genuinely split pair and 123 carry a non-tone `<ctcss>` body.

## Dependencies

None added.

## Validation

`npm test` passes in full (syntax, pyright, channels, webusb, settings, build).


web/js/datasources.js
```diff
@@ -348,15 +357,15 @@ export function buildPrzemiennikiRows(
...
       if (toneMode) {
         setRowValue(row, "Tone", toneMode);
       }
-      setRowValue(row, "rToneFreq", repeater.ctcssTx);
+      setRowValue(row, "rToneFreq", transmitTone);
     }
-    if (repeater.ctcssRx) {
-      setRowValue(row, "cToneFreq", repeater.ctcssRx);
+    if (receiveTone) {
+      setRowValue(row, "cToneFreq", receiveTone);
```
jasiek: [P2] Activate the normalized receive tone in the CHIRP mode

Writing receiveTone only to cToneFreq does not make the radio use it. CHIRP split_tone_encode() ignores cToneFreq when Tone is blank, and when Tone is Tone it encodes only rToneFreq. Therefore the new receive-only case produces no RX tone, and the unequal-pair case silently drops its receive side; live przemienniki.net records such as SR7W and SR8NP exercise the receive-only shape. Please derive the mode from both normalized values (TX-only: Tone; equal pair: TSQL; RX-only or unequal pair: Cross with ->Tone or Tone->Tone when supported, plus an explicit fallback for radios without Cross) and assert Tone/CrossMode in these tests.

jasiek: Fixed in d5e2379. You were right on the mechanism and on both records — I read `split_tone_encode` and it does exactly what you describe, so the receive-only case produced no RX tone and the unequal pair silently dropped its receive half.

`applyTonePair` now mirrors `chirp_common.split_tone_decode`, which is CHIRP's own inverse of the same mapping:

- transmit-only → `Tone`, rToneFreq
- matching pair → `TSQL`, cToneFreq
- anything else, receive-only included → `Cross` with `Tone->Tone` or `->Tone`

Each branch writes only the field its mode encodes, as `split_tone_decode` does.

On the Cross gate: `valid_cross_modes` defaults to the full `CROSS_MODES` list even when `has_cross` is false, so the CrossMode options are not evidence the radio can hold a split pair — the gate is the Tone column's own options. Fallbacks for a radio without Cross: a split pair keeps the access tone (drop it and the channel does not work at all), and a receive-only record takes TSQL, which costs an unrequested transmit tone but delivers the advertised squelch. Both are pinned by tests.

One thing I hit while checking your records against the live feeds, worth flagging separately: `<ctcss>` is not always a tone. Across 1,272 live US 2m RepeaterBook records it carries `CSQ` (44), `Restricted` (44) and DTCS codes such as `D023`/`D331` (35). `setRowValue` rejects those against the rToneFreq enum, so the mode logic would have set a tone mode backed by the row's default 88.5 — a channel transmitting a tone the directory never named. `parseCtcssFreq` now admits only a positive numeric body. Carrying the DTCS codes properly needs the DtcsCode/RxDtcsCode/DtcsPolarity columns; I left that out of scope.

The live pull also settles the premise the issue could only infer: przemienniki.net declares `repeater`, and SR4X carries qrg rx 145.000 / tx 145.600 with a lone `<ctcss type="rx">67.0`, so rx is the repeater's receive side on both elements. Note RepeaterBook declares `radio`, not `repeater` — my original PR description had that wrong; the code reads it per response, so no behaviour change.

PR #139, opened 2026-09-06
Report a runtime crash only when the bootstrap actually failed

### LINKED ISSUE #99: onRuntimeCrash fires on a heuristic and a failed bootstrap is never retried
## Summary

`onRuntimeCrash` is triggered by `!pyodide` rather than by an actual bootstrap failure. If the static catalog is missing and the jsDelivr index fetch in `listDriverModules()` fails before `loadPyodide` runs, it is logged as "RUNTIME CRASH" and `bootstrapFailed` latches. Conversely a failure inside `seedPyodideRuntime` (after `pyodide` is assigned) is never reported as a crash, and the rejected `bootstrapPromise` is never reset, so every later call re-fails until reload.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/runtime-rpc.js#L486-L489

## Reproduction

Block `cdn.jsdelivr.net` in DevTools and reload with no `radio-catalog.json`. Expected: a network error message. Actual: "RUNTIME CRASH" and no recovery without reload.

## Priority

Low.

jasiek:
Fixes #99.

## The bug

`onRuntimeCrash` fired on `!bootstrapFailed && !pyodide` — using "the pyodide handle is unset" as a stand-in for "the bootstrap failed", when it actually means "the bootstrap has not assigned yet". That proxy was wrong in **both** directions, and each direction hurt a different consumer:

- **False positive.** `listDriverModules()` is a plain HTTP fetch of `data.jsdelivr.com` and runs *before* `ensurePyodide()` in `loadRadioCatalogFromSources()`. With no `radio-catalog.json` and a blocked CDN, a network error was shown as `RUNTIME CRASH` and filed a false `error_kind: "runtime_crash"` event in Sentry.
- **False negative.** `ensurePyodide()` assigned `pyodide` from `loadPyodide()` *before* `seedPyodideRuntime()`. The genuinely reportable failure — the interpreter loaded but the runtime bridge never seeded — left the handle truthy, so it was never reported at all.
- **They compounded.** `bootstrapFailed` was a one-way, session-wide latch, so a spurious early network failure permanently silenced the report for a real crash later.
- **No recovery.** A rejected `bootstrapPromise` was cached for the life of the page; every later call re-awaited the same rejection until reload.

Net effect on the reporting added in 88d7f1e: false `runtime_crash` events from ordinary CDN blips, and silence for the failure mode most worth alerting on.

## The fix

New `web/js/runtime-bootstrap.mjs` owns the bootstrap sequence and replaces the heuristic with an explicit signal:

- Marks the errors it raises, so `runtime-rpc.js` can ask `isBootstrapFailure(error)` instead of inspecting an unrelated variable. The marker lives in a module-level `WeakSet` rather than a property or a `Symbol` key — this error object is handed to `captureError` and serialized into a Sentry payload, so a property would ride along as noise and a `Symbol` would vanish through `JSON.stringify`.
- Publishes the runtime handle only after seeding succeeds, so a half-built interpreter is never observable.
- Drops a rejected attempt instead of caching it, matching the reset-on-failure idiom `allDriverModulesPromise` already used a few lines above.
- `bootstrapCrashReported` is cleared on a successful bootstrap, making it a per-failure guard rather than a per-session one.

It is a separate module because `runtime-rpc.js` imports Pyodide from a CDN at parse time and therefore cannot be imported by the Node suite — which is why this code had no coverage. The new module is dependency-injected in the same style as `python-sources.mjs`.

## User-visible behaviour

A blocked or offline CDN now reads as a network error and can be retried without reloading the page. A genuinely broken runtime is reported instead of being swallowed.

## Tests

`scripts/test-runtime-bootstrap.mjs` (9 tests, registered in `test:channels`) covers the three previously untested cases plus concurrency: a pre-boot error is not a bootstrap failure; a seed failure is one and leaves the runtime unpublished; a failed attempt is retried rather than cached; concurrent callers share one attempt and one failure; a non-Error throw is still classifiable.

Full `npm test` passes (syntax, pyright, all suites).

## New dependencies

None.

## Notes

`FINDINGS.md` gains the crash-classification entry, and the existing `build-dist-rewrites-are-textual` finding is extended with something confirmed while checking that this new file packages correctly: `.mjs` is in neither `HASHED_EXTS` nor `REWRITE_EXTS` in `scripts/build-dist.mjs`, so the shared logic modules ship unhashed and unrewritten. Harmless today only because every `.mjs` here is a leaf that imports nothing — flagged rather than changed, as it is pre-existing and out of scope for this fix.


web/js/runtime-rpc.js
```diff
@@ -273,16 +277,22 @@ async function loadRadioCatalog() {
...
+    await seedPyodideRuntime(loaded, pythonSource);
+    return loaded;
+  },
+});
 
-  return bootstrapPromise;
+async function ensurePyodide() {
+  pyodide = await runtimeBootstrap.ensure();
+  // A working runtime ends the failure episode, so a later crash reports again.
+  bootstrapCrashReported = false;
```
jasiek: The reset here cannot make this a per-failure guard. Once ensure() succeeds, the bootstrap caches the runtime forever, so no later bootstrap attempt can fail; when a retry rejects, execution never reaches this assignment. For example, a first loadPyodide failure followed by a seedPyodideRuntime failure will report only the first error and silently suppress the distinct second failure until reload. Please scope deduplication to an attempt/error identity (or reset when a new attempt starts), and cover that sequence in a test.

web/js/runtime-rpc.js
```diff
@@ -535,8 +545,13 @@ export function createRuntimeRpcClient({
...
-        if (!bootstrapFailed && !pyodide && onRuntimeCrash) {
-          bootstrapFailed = true;
+        // Only the bootstrap itself is a runtime crash. This used to test the
+        // unset pyodide handle, which stood in for "bootstrap failed" but in
+        // fact meant "not assigned yet" -- so a failed driver-index fetch
+        // (plain HTTP, no Pyodide involved) was reported as a crash, and a
+        // failure seeding the bridge was not reported at all.
+        if (isBootstrapFailure(error) && !bootstrapCrashReported && onRuntimeCrash) {
+          bootstrapCrashReported = true;
           onRuntimeCrash(detailedError);
```
jasiek: This now captures the same bootstrap failure twice. ui.onRuntimeCrash() calls captureError(), then line 560 rethrows into the calling UI action, whose catch calls reportActionError() and captureError() again. Newly recognized seed failures therefore create duplicate Sentry events (and RUNTIME CRASH, RUNTIME ERROR, plus the action-error log). Since this failure returns through the normal action path, please propagate its classification and capture it once, or suppress the later action capture for an already-reported bootstrap error.

package.json
```diff
@@ -16,7 +16,7 @@
...
...ode --test --experimental-wasm-stack-switching scripts/test-channel-list.mjs scripts/test-ui-radio-loading.mjs scripts/test-call-queue.mjs scripts/test-runtime-bootstrap.mjs scripts/test-driver-import-race.mjs scripts/test-clipboard.mjs scripts/test-ui-channel-cut.mjs scripts/test-ui-channel-locations.mjs scripts/test-export-filename.mjs scripts/test-image-metadata.mjs scripts/test-datasources.mjs scripts/test-staticmap.mjs scripts/test-ui-repeater-map.mjs scripts/test-ui-repeater-query.mjs scripts/test-ui-repeater-modal.mjs scripts/test-ui-query-fields.mjs scripts/test-rsgb.mjs scripts/test-ui-file-drop.mjs scripts/test-dom-selectors.mjs scripts/test-debug-log.mjs scripts/test-analytics.mjs scripts/test-analytics-events.mjs scripts/test-sentry.mjs scripts/test-ga-dimensions.mjs scripts/test-gettext-builtins.mjs scripts/test-issue-report.mjs scripts/test-csv-export-parity.mjs scripts/test-power-level-watts.mjs scripts/test-channel-round-trip.mjs scripts/test-pyserial-shim.mjs scripts/test-serial-in-waiting.mjs scripts/test-metadataless-image-load.mjs scripts/test-progress.mjs scripts/test-clone-progress.mjs scripts/test-clone-button-gating.mjs scripts/test-clone-baud-plumbing.mjs",
```
jasiek: This line is currently conflicted with master. When resolving it, please retain scripts/test-chirp-import-errors.mjs and scripts/test-clone-detect-from-serial.mjs, which were added after this branch point, alongside the new runtime-bootstrap test. Taking this side as-is would silently remove those current suites from npm test.

web/js/runtime-rpc.js
jasiek: You're right, and the reset was worse than redundant — it was unreachable in every case that mattered. On a rejected retry the `await` in `ensurePyodide` throws before the assignment, and on success `ensure()` caches the runtime so no later bootstrap can fail. It could never re-arm.

Fixed in 79393b2 by keying deduplication on the failure rather than on a latch. `createBootstrapCrashReporter` in `web/js/runtime-bootstrap.mjs` remembers the last error object it reported: every call queued behind one attempt rejects with the *same* error object, so error identity is exactly the attempt — it collapses the queue into one report while still reporting the next attempt's different fault. `bootstrapCrashReported` and the reset are both gone.

Your exact sequence is now a test: "a second, distinct bootstrap failure is reported rather than swallowed" drives a `loadPyodide` failure, then a `seedPyodideRuntime` failure on the retry, and asserts both are reported. "every call queued behind one failed attempt reports that failure once" pins the other half, so the dedup can't regress into reporting per queued call.

web/js/runtime-rpc.js
jasiek: Confirmed — and it cut against the invariant FINDINGS already records, that `reportActionError` is the one capture funnel. Recognising seed failures newly routed them through both.

Fixed in 79393b2 with your second option, suppressing the later capture. `markBootstrapFailure` is now exported, and the runtime marks the fresh `Error` it rethrows, so the classification survives the `new Error(detailedError)` that would otherwise drop it. `reportActionError` returns before `captureError` for a marked error, keeping its log line and status — the user still needs to see which action died — while the single capture stays at `onRuntimeCrash`, which is the only place that can produce the `runtime_crash` tag.

I also dropped the redundant `RUNTIME ERROR` line when the same detail has just been logged as `RUNTIME CRASH`, so the panel shows the traceback once rather than twice.

Two tests in `test-debug-log.mjs` against a mock SDK: a marked bootstrap failure through `reportActionError` captures zero events, and an ordinary failure still captures one — the second is what stops the suppression widening into swallowing real action errors.

package.json
jasiek: Good catch — taking my side wholesale would have silently dropped both suites from `npm test`, which is exactly the kind of removal a green build would not have flagged.

Resolved in 63c4484 by taking master's line and inserting only `scripts/test-runtime-bootstrap.mjs` after `test-call-queue.mjs`. `test:channels` now runs 38 suites: master's 37, including `test-chirp-import-errors.mjs` and `test-clone-detect-from-serial.mjs`, plus the new one. I asserted the count and the three names by parsing the resolved `package.json` rather than eyeballing the line.

PR #140, opened 2026-09-06
Report memories the driver cannot decode instead of dropping them silently

### LINKED ISSUE #82: Swallowed get_memory failures on download turn into channel erasures on upload
## Summary

`_radio_rows_from_instance` does `except Exception: continue` around `radio.get_memory(number)`, so a memory the driver cannot decode silently disappears from the grid with nothing in the debug panel. `_apply_rows_to_radio_instance` then calls `erase_memory` on every valid number that is not present in the rows. One corrupt or unusual memory on the radio therefore vanishes on download and is wiped on the next upload. CHIRP shows such rows as errors and never auto-erases them.

#30 already asks for the swallow to be removed for the sub-device case; this issue is about the upload consequence for any radio.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L1006-L1014
- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L1060-L1061

Yes. Both loops re-read. Not reproduced at runtime.

## Reproduction

Requires a radio (or `.img`) with one memory the driver raises on: load it, note the row count is one short with no error logged, Upload (or Export Binary), reload the result. Expected: the undecodable memory is left untouched or reported. Actual: it is erased.

## Suggested fix

Log the traceback to the debug panel, keep a placeholder row (or a per-number "unreadable" marker) and never erase numbers that were unreadable on download.

## Priority

High: silent data loss on the radio.

jasiek:
Fixes #82.

## What was wrong

`_radio_rows_from_instance` skipped any memory `get_memory` raised on, with nothing written to the debug panel. The channel disappeared from the grid and the user had no way to know it existed.

That silence was destructive because the two loops are coupled through absence: `_apply_rows_to_radio_instance` treats every valid number *missing from the rows* as a user deletion. One undecodable memory therefore vanished on download and was erased on the radio at the next upload.

The erase half turned out to be already fixed. PR #123 (issue #83, immutable-policy validation) made the erase loop read each slot before erasing it and skip what it cannot read, which closed the data-loss path as a side effect on the same day #82 was filed. The download half was still silent, which is what this PR fixes. The line numbers cited in the issue had also shifted (1006/1060 to 1620/1691).

## What changed

- `_radio_rows_from_instance` logs the traceback to the debug panel and returns the numbers it could not decode alongside the rows.
- Both download paths — serial download and `.img` load — surface those as `unreadableChannels`.
- The status line names the count and points at Debug Output for the detail, so the missing rows are not silent in the UI either. Detailed diagnostics stay in the debug panel, per the repo's status-line convention.
- Failures are grouped by reason and the numbers capped per line, so a driver that raises on every slot in a 1000-channel `memory_bounds` cannot flood the panel. The existing erase-loop summary now shares that `_format_channel_numbers` helper, so both messages read the same way.

Upstream CHIRP takes the other option the issue suggested — `wxui/memedit.py` passes the exception itself into `_refresh_memory` and renders an error row. This PR uses the per-number marker instead: a placeholder row would have to survive `normalize_rows` and the upload path without being mistaken for something writable, which is a larger change than the reported defect warrants.

## Testing

New `scripts/test-unreadable-channels.mjs` drives the real `uv5r` driver against a real upstream image with `get_memory` forced to fail on one channel, wrapping `erase_memory` rather than stubbing it so the assertion covers what the driver was actually asked to do. It pins both halves: the download reports the failure and marks the number, and the upload leaves the slot alone.

The test was checked against the bug rather than only against the fix — restoring the pre-#123 erase loop makes it fail with `channel 0 was erased despite being unreadable`.

Registered in `test:channels`. Full suite passes (1062 tests), `pyright` reports 0 errors. Three call sites in `test-channel-list.mjs` were updated for the new tuple return.

## Dependencies

None added.


web/python/runtime_bridge.py
```diff
@@ -2128,14 +2161,15 @@ def _download_selected_radio_sync(module_name: str, class_name: str):
...
-    rows = _radio_rows_from_instance(radio)
+    rows, unreadable = _radio_rows_from_instance(radio)
     csv_text = normalize_rows(rows, module_name, class_name)
     settings_result = _validate_and_apply_radio_settings(radio, [], apply_changes=False)
     return {
         "rows": rows,
         "headers": CSV_HEADERS,
         "csvText": csv_text,
         "settings": settings_result["settings"],
+        "unreadableChannels": unreadable,
```
jasiek, line 2262: [P1] Preserve this set through upload and export

This value is only returned for display. Neither uploadSelectedRadio nor exportImage sends it back, and both Python paths reconstruct a fresh radio and call _apply_rows_to_radio_instance with only the rows. If the initial get_memory failure is one-shot or instance-local, the fresh read succeeds and the omitted slot is erased. I reproduced this with two real UV-5R instances: extraction returned unreadable [0], then apply on the fresh instance called erase_memory(0). Please persist/pass protected locations and subtract them from the erase candidates; make the regression test restore get_memory or use a fresh instance before applying.

web/python/runtime_bridge.py
```diff
@@ -1617,21 +1618,56 @@ def _iter_memory_numbers(radio):
...
     rows: Rows = []
+    unreadable: list[int] = []
+    numbers_by_reason: dict[str, list[int]] = {}
+    trace_by_reason: dict[str, str] = {}
     for number in _iter_memory_numbers(radio):
         try:
             mem = radio.get_memory(number)
-        except Exception:
+        except Exception as exc:
+            reason = str(exc) or exc.__class__.__name__
```
jasiek, line 1730: [P2] Bound the number of failure groups, not just numbers per group

The complete exception text is the grouping key. If a driver includes the memory number or invalid value in the message, every slot gets a unique key, and the loop below logs one full traceback per slot. The eight-number per-line cap therefore does not protect the debug textarea in that case. Please cap distinct groups/tracebacks globally (and summarize the remainder), or normalize the grouping key, and add a test with unique per-channel messages.

jasiek (commented):
Found two actionable issues below. The first leaves issue #82's data-loss path open when a decode failure is transient or instance-specific; I reproduced channel 0 being returned as unreadable during extraction and then erased by apply on a fresh UV-5R instance. The second means unique per-channel error messages can still flood Debug Output. The targeted unreadable-channel tests, JavaScript syntax checks, and pyright all pass. The branch is also currently merge-conflicting with master in FINDINGS.md, package.json, and web/python/runtime_bridge.py.

web/python/runtime_bridge.py
jasiek: Confirmed and fixed in bac21ab. You were right that returning the set for display only left the path open — I had flagged the transient case in the handover but stopped short of closing it, which was the wrong call given it is the actual data-loss vector.

Extraction now records the numbers in UNREADABLE_BY_DRIVER, keyed with _driver_cache_key exactly like LAST_IMAGE_BY_DRIVER so the protection resolves wherever the cached image does, and _apply_rows_to_radio_instance subtracts them from the erase candidates. Both download paths write it, and the write always overwrites (including with an empty set) so a later clean read of the same radio drops protection that no longer applies rather than leaving a slot un-erasable for the session.

I kept it server-side rather than round-tripping through uploadSelectedRadio and exportImage: the JS would have to carry it through the grid, merge, and CSV round-trip without ever dropping it, and a drop there is silent data loss again. Keying it to the cached image means the protection has the same lifetime as the bytes it protects.

A row the user does supply for a protected number still writes normally — explicit intent lands in seen_numbers and never reaches the erase loop.

The test now does what you asked, using the class-level patch so it matches your reproduction: the failure is injected on the driver class, lifted before the export, and the export therefore reads the slot without error on its fresh instance. Rather than asserting erase_memory was not called, it re-loads the exported image and checks the channel is still there with its original frequency. Verified against the defect — without the subtraction it fails with "channel 0 did not survive the export intact".

web/python/runtime_bridge.py
jasiek: Correct, and the numbers are worse than I assumed — a UV-5R failing on every slot produced 128 tracebacks, because the driver message names the memory and the grouping key is the message.

Fixed in bac21ab by capping distinct groups with MAX_LOGGED_FAILURE_GROUPS (5) rather than normalising the key. Normalising would have meant guessing at driver message formats to strip the variable parts, and getting that wrong merges genuinely different faults into one line; capping is exact and loses nothing that matters, since the untruncated groups are still counted and their channels still named in a summary line.

Reporting is ordered by descending channel count, so truncation drops the long tail of one-off messages instead of the failure that explains the most channels — with unique-per-slot messages every group is size 1 and the tie breaks on the message text, but in the mixed case the widest fault is always reported.

Both reporting sites now share _log_grouped_channel_failures, so the erase-loop summary is bounded the same way; it had the same unbounded-groups shape.

New test drives unique per-channel messages through load_image_base64 and pins the traceback count to the cap plus exactly one summary line. It fails against the unbounded version with "expected at most 5 tracebacks, got 128".

PR #147, opened 2026-09-07
Share test fixtures via scripts/test-support and memoize the Pyodide harness

jasiek:
## Summary

The node:test suite under `scripts/` had grown by copy-paste: twelve files each carried their own `FakeElement`, four WebUSB chip tests their own fake `USBDevice`, nine files an identical `setNavigator`, fifteen a local `repoRoot`, and two 160-line blocks were byte-identical. This PR introduces `scripts/test-support/` and moves every shared fixture there. No test assertion or coverage changes; the suite is ~1,100 lines smaller net and a little faster.

## What changed

- **`test-support/fake-dom.mjs`** replaces the per-file fake DOM classes. Bespoke behaviours (progress-bar value reflection, strict selector matching, focus logging) survive as options or small local subclasses with a comment saying why.
- **`test-support/fake-usb.mjs` and `usb-read-path.mjs`** hold the shared USB device skeleton and the three read-path tests that were copied into all four chip files. Each chip file keeps its prose comments and a thin wrapper supplying only its control-transfer answer table.
- **`test-support/fake-serial.mjs` and `globals.mjs`** provide one recording serial port for the port-loss, reconfigure, baud-rate, control-line and in-waiting tests, plus a navigator stub that restores itself in `t.after` instead of leaking.
- **`test-support/chirp.mjs`** wraps the Pyodide harness with the repeated catalog, image and Python-snippet helpers. `createTestRadioHarness()` now memoizes the booted runtime per option set (each boot cost ~1.3 s and six files booted once per test); tests that mutate Python state pass `isolated: true`. `test-channel-list.mjs` no longer carries its own copy of the harness bootstrap.
- **`fake-window.mjs`, `row-hooks.mjs`, `temp-dir.mjs`, `repo-paths.mjs`** cover the analytics/Sentry window, the four `rowHooks` copies, temp-dir cleanup, and path constants that no longer depend on the invocation directory.
- **`check:syntax` now walks `scripts/` too.** It only covered `web/`, so the test scripts themselves were never parse-checked before commit.
- **FINDINGS.md** gains a Test suite section documenting the convention, the harness memoization and its escape hatch, and the one intentional behaviour change below.

## Reviewer notes

- One deliberate fake-behaviour change: the old per-file `<select>` fakes skipped a placeholder option with value `""` and answered with the next one. The shared `FakeElement` returns the first option's value verbatim, as a browser does. No existing assertion depended on the old shortcut.
- `origin/master` was merged in; the no-op export test from #142 and the memory-comments test from #143 were moved onto the shared helpers rather than keeping fresh copies of the removed ones.
- Full `npm test` passes: 363 + 161 + 561 + 11 tests, 0 failures. Round-trip file runtime dropped from 12.9 s to 9.3 s.
- Some commits were authored in parallel worktrees per cluster and merged; the history is linear per cluster and small per commit.

## Dependencies

No new dependencies.


jasiek:
@codex review

PR #148, opened 2026-09-07
Prevent offline exports from enabling unsafe uploads

### LINKED ISSUE #74: Offline Export Binary fabricates an all-zero cached image, letting Upload bypass the "no downloaded image" gate
## Summary

`export_image_base64` falls back to `bytes(memsize)` when nothing has been downloaded, and then stores the resulting packed image in `LAST_IMAGE_BY_DRIVER`. `_upload_selected_radio_sync` only checks that a cached entry exists, so after one offline export an Upload proceeds with zeroed settings, ident and calibration regions as its base image. `get_radio_settings` also lights up with values from the zeroed image. This violates the CLAUDE.md rule that upload must fail clearly when no cached image exists.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L1377-L1399 (fallback and `LAST_IMAGE_BY_DRIVER[driver_key] = bytes(packed)`)
- Gate that is bypassed: https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L1301-L1304

Yes. Cited lines re-read. Not reproduced at runtime.

## Reproduction

1. Select any clone-mode radio without connecting.
2. Import a CSV (or type a few channels).
3. Click Export Binary. It succeeds.
4. Connect the radio and click Upload.

Expected: "no cached image, download first". Actual: upload runs against a synthetic all-zero image.

## Suggested fix

Do not store the synthetic image in the cache, or store it under a separate key that the upload gate rejects.

## Priority

High: can write zeroed calibration and settings regions to a radio.

jasiek:
## Summary

- keep offline binary exports from populating the trusted radio-image cache
- preserve cache updates when an export starts from a downloaded or loaded image
- add regression coverage that verifies export still works while settings and upload remain image-gated

Closes #74

## Testing

- node --test --experimental-wasm-stack-switching scripts/test-channel-list.mjs
- npm run check:syntax
- npm run check:types
- npm run test:channels
- npm run test:webusb
- npm run test:settings
- npm run test:build
- python3 -m py_compile web/python/runtime_bridge.py
- git diff --check

## Dependencies

No new dependencies.


jasiek:
@claude review this

jasiek:
@claude review this

PR #149, opened 2026-09-07
Skip settings CHIRP could not load instead of failing the upload

### LINKED ISSUE #75: WebSerialPipe.baudrate is a dead attribute: 32 drivers that switch speed mid-clone cannot work
## Summary

`WebSerialPipe.baudrate` is a plain instance attribute. Assignments by drivers never reach the JS bridge, so the port stays at the rate chosen at connect time. 32 upstream drivers change `pipe.baudrate` during a clone, including every Icom clone via `icf.py:328` (to 38400 after the handshake), `thd72.py:478` and `thd74.py:243` (57600), `nx800.py:118` and `tk8180.py:650-660` (19200 for the ack), `alinco.py:672`, `vgc.py:437` (115200). Their second phase reads garbage or times out.

Affected driver modules (from `grep -rl "pipe.baudrate *=" chirp/chirp/drivers`): ar8200 ga510 ft450d bf_t1 tk3140 ts590 tdxone_tdq8a thd72 tk760 kenwood_d7 hf90 thd74 nx800 baofeng_common icm710 ic9x icf tk280 lt725uv tk690 tmv71 vgc tmd710 alinco icomciv ts480 th_uv88 btech tk8180 hobbypcb ts790e th_uv8000.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L814-L820 (attribute only)
- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/python/runtime_bridge.py#L1075
- Web Serial only accepts `baudRate` at `open()`, so a fix needs a close/reopen path in `web/js/serial.js` triggered from a `baudrate` property setter.

Yes. Shim re-read; driver list confirmed by grep against the submodule. Not reproduced on hardware.

## Reproduction

Requires an affected radio, for example any Icom clone-mode model or a Kenwood TH-D72: connect, Download. Expected: clone completes. Actual: handshake succeeds, then timeout or garbage once the driver switches speed.

## Related

#39 covers the same gap for stop bits.

## Priority

High: whole radio families cannot clone.

jasiek:
Fixes #75.

## The bug

Uploading to a UV-5R aborted with:

```
File "/webchirp_runtime/chirp/drivers/uv5r.py", line 2156, in set_settings
  setattr(obj, setting, element.value)
File "/webchirp_runtime/chirp/bitwise.py", line 578, in set_value
  self._data[self._offset] = (int(value) & 0xFF)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

When a driver builds a setting from image content its own validation rejects, `RadioSettingGroup.__init__` logs the failure and swallows it (`chirp/settings.py:493-504`), leaving `_current` at None. The setting stays in the tree `get_settings()` returns.

Whether writing that tree back survives is a property of the **driver**, not the image. `ft4` tolerates it — its `cw_id`/`passwd` hold `'------'`, which its own charset rejects, on five upstream fixtures. `uv5r.set_settings()` writes every element straight through with `setattr(obj, setting, element.value)`, so the None reaches `int()` in `bitwise.py` and kills the whole upload.

This needs no corrupt codeplug. `uv5r` declares `abr` as 0-24 and `squelch` as 0-9 over full `u8` fields, so any radio storing a byte outside the range its driver declares aborts the upload — over a setting the user never touched.

The exception depends on the value type, so reports of this one bug do not look alike:

| Value type | Exception |
| --- | --- |
| `RadioSettingValueInteger` | `TypeError: int() argument ... not 'NoneType'` |
| `RadioSettingValueList` | `ValueError: None is not in list` |

## The fix

Handle it the way desktop CHIRP does. `_remove_dead_settings` (`chirp/wxui/settingsedit.py:177-190`) strips immutable **and** uninitialized settings from a deep copy before `set_settings`, so those values never reach a driver at all. `_prune_dead_settings()` now does the same in `_validate_and_apply_radio_settings`.

Pruning a copy is load-bearing rather than merely defensive: the same return statement serializes `settings_tree` back to the UI, so pruning in place would make dead settings vanish from the panel after an upload. Dropped settings are named in the debug panel.

Two details worth a reviewer's eye:

- `RadioSetting` subclasses `RadioSettingGroup`, so the isinstance order matters — recursing into a setting would walk its values, not settings.
- Three container shapes need two removal idioms: `RadioSettingGroup.__delitem__` takes the element, while a `RadioSettings` root and the bare list `icf520` returns are plain lists. Hence `_remove_settings_child`.

## Behaviour change to be aware of

Pruning now also drops **immutable** settings on ordinary images (`firmware_msg.line1`, `model`, `oem1`, `modelinfo.freqrange`, ...). That is the other half of desktop's rule, and `uv5r.set_settings` already guarded it itself — but for drivers that do not guard, this is a genuine change: those settings are no longer handed to the driver. It matches CHIRP, so it was kept.

## Tests

Added a case to `scripts/test-settings-round-trip.mjs` covering both exception shapes — `abr = 255` (integer) and `timeout = 0x7f` (list). No upstream image carries both halves of the bug (a naive driver *and* a rejected value), so the byte is planted. The test asserts the planted byte genuinely reads back uninitialized, so it cannot pass vacuously; that the write succeeds; and that `after === before`, because a skipped setting must not be repaired or zeroed behind the user's back.

Verified the test fails without the fix with the exact reported `TypeError`.

Full suite: 1101 tests, 0 failures. `pyright` and `py_compile` clean.

## Dependencies

None added. `copy` is stdlib.

## Docs

`FINDINGS.md` gains a `dead-settings-must-be-pruned-before-write` entry, and the settings round-trip summary line now mentions the pruning.


jasiek:
@codex review

PR #150, opened 2026-09-07
Guard the repeater query form against a second in-flight submit

### LINKED ISSUE #105: Repeater query form has no in-flight guard: double-clicking Query API inserts every repeater twice
## Summary

The form `submit` handler in `repeater-query.js` has no busy flag and never disables the submit button while `runQuery` is awaiting. `dom.repeaterQuerySubmitEl` is declared in `dom.js` but never referenced. An RSGB fan-out over 24 squares takes a few seconds; clicking "Query API" twice resolves both queries and each calls `insertRowsAtSelectionOrEnd`, duplicating every repeater. Same for przemienniki, RepeaterBook and IRTS.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/repeater-query.js#L196-L209
- Unused element: https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/dom.js#L81

Yes. Handler re-read; grep confirms `repeaterQuerySubmitEl` is unused outside `dom.js`. Not reproduced at runtime.

## Reproduction

1. Channel menu, "Query RSGB ETCC", enter locator IO82MM with a large radius.
2. Double-click "Query API".

Expected: one set of rows. Actual: every repeater appears twice.

## Priority

Medium.

jasiek:
Fixes https://github.com/jasiek/webchirp/issues/105.

## The bug

`repeater-query.js`'s form `submit` listener is `async`, so `await activeSource.runQuery(...)` returns control to the event loop. Nothing stopped a second `submit` from re-entering the handler while the first was suspended: both queries resolved, both called `insertRowsAtSelectionOrEnd`, and every repeater landed twice. An RSGB fan-out over 24 locator squares takes seconds, so double-clicking "Query API" hits it easily. The same applied to przemienniki, RepeaterBook and IRTS, since all four share the one modal.

## The fix

- A module-scoped `queryInFlight` flag rejects a re-entrant submit. The flag, not the button state, is the authoritative check — pressing Enter in a text field submits the form too.
- `dom.repeaterQuerySubmitEl` (declared in `dom.js` but never referenced until now) is disabled and relabelled "Querying..." while the query runs, so the user can see why a second click does nothing.
- The reset lives in a `finally`, so a failed, rejected or browser-timed-out query leaves the form usable for a retry instead of wedged. The modal already stays open on an error, so retrying is in place.

## Tests

- New test in `scripts/test-ui-repeater-modal.mjs`: a gated fetch holds a query in flight while a second submit is dispatched, then asserts one insert, one fan-out, and the button restored to "Query API". Verified it fails against the unfixed handler (two inserts) and passes with the guard.
- The existing "coordinate-free or ill-formed RSGB query" test now also asserts the button is re-enabled after a failed submit, covering the `finally` path.
- Full `npm test` green: 1105 tests, 0 failures, plus the syntax and pyright checks.

## Dependencies

None added.

## Noted but out of scope

None of the three `fetch` call sites in this path use an `AbortController`/`AbortSignal.timeout`, so there is no application-level timeout — a hung proxy leaves the button reading "Querying..." until the browser's own network timeout fires. Pre-existing, and worth its own change.

### DIFF
--- a/scripts/test-ui-repeater-modal.mjs
+++ b/scripts/test-ui-repeater-modal.mjs
@@ -740,6 +740,32 @@ function installRsgbFetch(bySquare) {
   return calls;
 }
 
+// Like installRsgbFetch, but every response waits on a gate the test opens.
+// That holds a query in flight for as long as the test needs, which is what it
+// takes to land a second submit on the first one — the double-click the
+// in-flight guard exists to absorb.
+function installGatedRsgbFetch(bySquare) {
+  const calls = [];
+  let openGate;
+  const gate = new Promise((resolve) => {
+    openGate = resolve;
+  });
+  Object.defineProperty(globalThis, "fetch", {
+    configurable: true,
+    value: async (url, init) => {
+      calls.push({ url: String(url), init });
+      await gate;
+      const square = String(url).split("/").pop();
+      return {
+        ok: true,
+        status: 200,
+        json: async () => ({ data: bySquare[square] ?? null }),
+      };
+    },
+  });
+  return { calls, release: () => openGate() };
+}
+
 function repeaterRecord(overrides = {}) {
   return {
     id: 199,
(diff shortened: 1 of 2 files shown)


jasiek:
@codex review

PR #151, opened 2026-09-07
Bound repeater directory queries with a 10 s request deadline

jasiek:
## Problem

None of the online repeater directory fetches had an application-level timeout. A proxy or API that accepts the connection and then hangs left the query modal's submit button disabled reading "Querying..." with nothing in the debug panel, until the browser's own network timeout fired — roughly 300 s in Chrome for a stalled connection, and not configurable from the page.

Affected call sites:

- `web/js/ui/repeater-sources.js` — the `/meta` dictionary fetch (przemienniki.net, RepeaterBook, IRTS)
- `web/js/ui/repeater-sources.js` — the RXF query fetch for the same three sources
- `web/js/rsgb.js` — `fetchRsgbRecords`, which fans out over locator squares under `Promise.all`

## Change

New `web/js/request-timeout.js` holds the single deadline (`REPEATER_REQUEST_TIMEOUT_MS` = 10 s) and `withRequestTimeout(label, run, timeoutMs)`. All three call sites go through it. A stall now rejects with, for example:

> `przemienniki.net query timed out after 10 s: the server accepted the request but never answered.`

which reaches `log.reportActionError` and so lands in the debug panel and the status line, and matches the existing `timeout` pattern in `classifyErrorKind` rather than falling through to `other`.

Three design points worth reviewing:

- **The deadline wraps the body read, not just the fetch.** `fetch()` resolves as soon as the response *headers* arrive, so a host that answers `200` and then stalls mid-body would slip past a request-only deadline. Hence the `run(signal)` callback shape, which keeps `response.text()`/`.json()` inside the window. The error path is inside it too — reading the body of a failed response can stall exactly as the success path can.
- **`AbortSignal`, not `Promise.race`.** A race leaves the connection open and still streaming; only an abort tears it down. It also adds nothing to the wire: `web/js/rsgb.js:277` documents that any custom header would force a CORS preflight the RSGB API answers with 405, and a signal is not a header. **No request headers were introduced.** The existing preflight regression assertions (`call.init === undefined`) were tightened rather than dropped — they now assert `init` carries the key `signal` and nothing else.
- **Our own `AbortController`, not `AbortSignal.timeout()`.** `signal.aborted` is then unambiguous proof that *our* deadline fired, so a caller's own error (`HTTP 502`) keeps its message while a genuine timeout gets a readable sentence instead of the platform's "signal is aborted without reason".

### RSGB fan-out decision

**One timed-out square fails the whole query** — the same as any other error there already did, and kept that way deliberately. A locator square is a geographic tile, so skipping a stalled one would hand back a result set with an invisible hole in the middle of the search area, and the user would write those channels to a radio believing the missing repeaters simply are not there. A failed query they can retry is the honest outcome, and the error names the square. The deadline is *per square*, so the fan-out stays bounded by one timeout rather than by their sum. This is documented in the comment above `fetchRsgbRecords` and in FINDINGS.md.

## Tests

`scripts/test-ui-repeater-modal.mjs` — four end-to-end tests over `t.mock.timers`, so they assert the shipped constant and its exact user-facing message rather than a value only the test knows. A new `installGatedFetch` helper routes a request to a promise that settles only when the deadline aborts its signal, which is what a hung proxy looks like from the browser.

- a stalled RSGB request times out, names the square, inserts nothing and leaves the modal open with the user's filters
- a stalled dictionary request times out instead of leaving the modal opening forever
- a stalled query times out with the dictionary already loaded
- a request answering inside the deadline is unaffected, and its timer is cleared

`scripts/test-rsgb.mjs` — four unit tests using the new `timeoutMs` option: the abort actually reaches the fetch (not just an abandoned promise), the deadline covers a stalled *body read*, one stalled square fails the whole fan-out, and a square answering in time is untouched.

`npm test` passes (1112 tests, exit 0).

## Dependencies

None added.


jasiek:
@codex review

jasiek:
Rebased onto master now that #150 has merged. Three things changed in the reconciliation:

- **Conflict resolved** in `scripts/test-ui-repeater-modal.mjs` — both branches appended tests to the end of the file; both test blocks are kept.
- **Helper renamed** `installGatedFetch` → `installStalledFetch`, to keep it apart from #150's `installGatedRsgbFetch`. They hold a request open for opposite reasons: #150's gate is opened *by the test*, so a query can be caught mid-flight; this one is never released at all, so only the deadline can end it. Cross-referenced in both comments.
- **New assertion tying the two features together**: a timed-out query now asserts the submit button comes back enabled and relabelled. #150's `finally` claims a timed-out query leaves the form usable rather than wedged, and until this branch there was no timeout to prove it with — a deadline that left the modal stuck on "Querying..." would just have swapped one stuck modal for another.

Also corrected the rationale comment for not using `AbortSignal.timeout()`, which overstated its case. The attribution argument was wrong: `AbortSignal.timeout()` aborts with a `TimeoutError` DOMException, so its `reason.name` distinguishes a deadline from any other abort just as well as `signal.aborted` does here. The real reason is testability, and it is verifiable:

```js
t.mock.timers.enable({ apis: ["setTimeout"] });
const viaTimeout = AbortSignal.timeout(10000);
const controller = new AbortController();
setTimeout(() => controller.abort(), 10000);
t.mock.timers.tick(10000);
// AbortSignal.timeout aborted after tick: false
// own AbortController aborted after tick: true
```

`AbortSignal.timeout()`'s timer is internal to the platform, so no test runner can advance it — a deadline built on it can only be tested by really waiting out the real duration, which makes the shipped value untestable and forces the tests to assert a short value production never uses. `setTimeout` + `AbortController` costs four lines and lets the four modal tests drive the real `REPEATER_REQUEST_TIMEOUT_MS` over a mocked clock. Secondary: it also keeps working on Chrome 89–102, which has Web Serial but not `AbortSignal.timeout` (Chrome 103).

`npm test` passes on the rebased branch (exit 0).

PR #153, opened 2026-09-08
Skip repeaters whose CTCSS tone the radio cannot send

### LINKED ISSUE #104: Repeater tones not in the radio's table silently become 67.0 Hz with Tone enabled
## Summary

`setRowValue(row, "rToneFreq", ctcss.toFixed(1))` goes through `normalizeValue`'s enum handling, which on no match returns `previous ?? options[0]`, typically "67.0", while Tone has already been set to "Tone". A directory record with a non-standard or mistyped tone, or a radio with a reduced tone table, is programmed with 67.0 Hz, which looks valid and keys nothing. Present in both the RSGB and przemienniki builders.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/rsgb.js#L513-L520
- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/datasources.js#L336-L341
- Fallback: `web/js/ui/channel-values.js:79-93`

Yes by static review (mechanism certain; real-world frequency of off-table tones is low). Not reproduced at runtime.

## Reproduction

Unit test: build an RSGB row with `ctcss: 12.3` against a standard tone table. Expected: Tone left blank or row skipped with a reason. Actual: Tone "Tone" with rToneFreq "67.0".

## Suggested fix

Check the return of `setRowValue` for the tone and clear Tone (or skip with a reason) when the value was not accepted.

## Priority

Medium.

---

## Triage 2026-09-07 (against 87c032e)

Still valid, and **wider than filed** — three unchecked write sites rather than one, and a third affected source.

The fallback is unchanged at `web/js/ui/channel-values.js:78-93`: `return String(previous ?? options[0] ?? "")`. The only added leniency is a numeric-equality retry (`:83-91`), which still misses tones genuinely absent from the driver's table. `previous` is never empty on a fresh row — `defaultValueForColumn` (`web/js/ui/channel-table.js:146-148`) seeds enum columns with `String(meta.options[0])`, typically `"67.0"` — so the fallback lands on the first tone either way.

**The structural reason no caller can check the write:** `setRowValue` is `setRowValueIfPresent` (`channel-table.js:270-276`, exported via `rowBuilderHooks()` at `:303`) and **returns nothing**.

Unchecked writes, each occurring after the Tone mode has already been committed:

- RSGB — `web/js/rsgb.js:515-519`.
- przemienniki / RXF — `web/js/datasources.js:154-201`, `applyTonePair`: `writeTransmitOnly` (`:157-162`), the Cross branch (`:195-200`) and `writeReceiveAsTsql` (`:164-170`). `parseCtcssFreq` (`:133-139`) rejects only non-numeric bodies such as "CSQ" and "D023", so a numeric but off-table tone like `141.3` passes straight through.
- **IRTS** routes through the same przemienniki builder (`web/js/ui/repeater-sources.js`), so it is affected too.

Not affected: `buildPmr446Rows`, `buildFrsRows` and `buildGmrsRows` set `Tone` to `""` (`datasources.js:333`, `:354`, `:389`).

Shares a root cause with #93 and #94. #120 already established that `setRowValue` is a validating writer rather than an assignment, but fixed only the band call site — giving it a return value would close all three.

jasiek:
Fixes #104.

## The bug

`setRowValue` is a validating writer, and a rejected **enum** write is invisible: `normalizeValue` falls back to `previous ?? options[0]`, and a fresh row already holds `options[0]` because `defaultValueForColumn` seeds enum columns with it. For a CTCSS table that first option is 67.0 on nearly every driver.

Both repeater builders committed the tone *mode* first and wrote the tone afterwards, so a repeater whose access tone the radio's table does not carry was imported as **Tone + 67.0 Hz** — a channel that looks correctly programmed and opens nothing. Confirmed the reduced tables are real, not hypothetical: `ft90` drops 165.5/171.3/177.3 from `TONES`, and the older Kenwoods use `OLD_TONES`, which drops eleven tones including 159.8, 177.3, 183.5 and 254.1.

Wider than filed, as the triage note said: RSGB (`web/js/rsgb.js`) plus all three RXF sources — przemienniki.net, RepeaterBook and IRTS — through `applyTonePair` in `web/js/datasources.js`, which had three unchecked write sites of its own.

## The fix

**Give the writer a signal.** `normalizeCellValue` returns `{ value, accepted }`; `normalizeValue` is kept as its value-only wrapper, so no existing call site changes. `setRowValue` returns that boolean. `accepted` is false on every path that falls back instead of storing the caller's value: a read-only column, an unparsable or out-of-band frequency, a clamped or non-numeric int, an enum the driver does not offer. Coercions that keep the value — text trimmed to `validChars`/`maxLength`, an enum matched numerically ("88.50" → "88.5") — stay accepted.

**Write the tone before the mode that encodes it**, and commit the mode only once the tone landed. A repeater whose *access* tone cannot be written is skipped with reason `"tone"` and reported like the existing frequency and mode skips: "skipped 1 needing a tone it cannot send" in the status line, and the frequency itself in the debug panel. A tone that can only be *received* costs the squelch and not the channel, so that row is kept with the tone mode left clear rather than dropped. Under Cross, both halves must land before the mode is committed, otherwise the row falls back to the single side the radio can hold.

A driver that publishes no tone table at all (`valid_tones = []`, e.g. ar8200, hf90, icm710) has an empty option list, which `normalizeValue` treats as "no constraint" — so those radios accept every tone and nothing is skipped.

## Tests

- `scripts/test-channel-values.mjs` (new, wired into `npm run test:channels`) pins the acceptance flag per metadata kind.
- `test-rsgb.mjs` and `test-datasources.mjs` cover an off-table access tone, a radio with no tone mode at all, a matching pair the table lacks, a receive-only tone (row kept, mode clear) and a split pair with one unwritable half.
- `test-ui-repeater-modal.mjs` covers the status line and debug wording end to end.
- The shared `scripts/test-support/row-hooks.mjs` now validates enum writes and reports acceptance, matching the grid.

Full suite green: 1136 tests, 0 failures, pyright clean. Also spot-checked against real CHIRP metadata for the UV-5R: 103.5 and 141.3 accepted, 12.3 rejected to 67.0 with `accepted: false`.

## Dependencies

None added.


jasiek:
@codex review

PR #155, opened 2026-09-08
Simplify the webchirp_bridge package

jasiek:
## Summary

Stacked on #154 (the package split); this PR only touches `web/python/webchirp_bridge/`. Eleven behaviour-preserving refactors, one commit each, taken from a read-only audit of the freshly split package. Every public function keeps its name, signature and JSON payload shape; the module graph stays acyclic. Net: 337 insertions, 408 deletions.

| Commit | What it removes |
| --- | --- |
| Use `_js_to_py` | two hand-rolled `hasattr(x, "to_py")` conversions |
| Remove `_normalize_power_value` | dead code, unreferenced in `web/`, `scripts/` and docs |
| Base64 helpers in `images.py` | three copies of the decode-or-raise block, two of the `imageBase64`/`size` pair |
| `_temp_image_path` context manager | three NamedTemporaryFile + swallowed-unlink dances |
| `_row_from_memory` / `_row_text_values` | four zips of `CSV_HEADERS` against `_row_values_for_csv` |
| `MEMORY_FIELD_HEADERS` | three independently maintained copies of the same sixteen-field map |
| `_level_map_for_radio` / `_blank_radio_instance` | duplicated power-level fallback and `radio_cls(None)`-then-`""` constructor |
| `_issue` / `_setting_issue` / `_settings_validation_payload` | nine hand-built issue dicts, three seven-key settings replies, two duplicated message strings |
| Column metadata builders | 21 hand-written column dicts in four recurring shapes; `DV_ONLY_HEADERS` now drives the D-STAR columns too |
| `_read_radio_payload` | the five-step post-read tail shared by serial download and image load |
| `_read_memory` / `_erase_memory` | four inline `ExternalMemoryProperties` guards |

## Deliberately not done

- `upload_image_base64` is a public entry point with no caller anywhere. Left in place; delete it if you agree it is dead.
- The cached-image preamble shared by upload and export was left alone: the two callers differ in cache-miss policy and progress reporting.
- Latent inconsistency, unchanged here: the upload path reads existing memories through `get_memory_extra` but the preflight in `validate_rows_for_upload` does not, so the two can disagree about a row that only changes a comment on an `ExternalMemoryProperties` radio.

## Verification

`npm test`: 396 + 161 + 562 + 18 passing after every commit's compile and pyright check, and on the final tree. `npm run check:python` clean.

## Dependencies

No new dependencies.

### DIFF
--- a/web/python/webchirp_bridge/channel_rows.py
+++ b/web/python/webchirp_bridge/channel_rows.py
@@ -10,7 +10,7 @@ under ``ROW_EXTRA_KEY`` and are applied to memories here as well.
 
 from __future__ import annotations
 
-from typing import Any, TYPE_CHECKING
+from typing import TYPE_CHECKING
 
 from chirp import (
     chirp_common,
@@ -28,23 +28,23 @@ from webchirp_bridge.power_levels import (
 from webchirp_bridge.runtime_errors import RuntimeUnsupportedError
 
 if TYPE_CHECKING:
-    from typing import Optional, Sequence
+    from typing import Any, Optional, Sequence
 
-# A channel as it crosses the JS/Python boundary: one JSON object per channel,
-# keyed by CSV header name (``CSV_HEADERS`` below, from
-# ``chirp_common.Memory.CSV_FORMAT``) with text values — "Location": "25",
-# "Frequency": "443.000000", "Duplex": "+". It is the grid's row, serialized by
-# ``setRowsJsonGlobal()`` in ``web/js/runtime-rpc.js`` and parsed here with
-# ``json.loads``, so every value a header names is a string.
-#
-# The value type is ``Any`` rather than ``str`` because a row may also carry
-# non-header keys the editor rides along on it — currently the ``__geo``
-# sidecar (``web/js/row-geo.js``), an object, which is why the type cannot
-# promise ``str`` for arbitrary keys. Nothing in the runtime reads those: every
-# consumer projects a row through ``CSV_HEADERS`` and ignores the rest, which is
-# what keeps the sidecar out of a codeplug.
-Row = dict[str, Any]
-Rows = list[Row]
+    # A channel as it crosses the JS/Python boundary: one JSON object per channel,
+    # keyed by CSV header name (``CSV_HEADERS`` below, from
+    # ``chirp_common.Memory.CSV_FORMAT``) with text values — "Location": "25",
+    # "Frequency": "443.000000", "Duplex": "+". It is the grid's row, serialized by
+    # ``setRowsJsonGlobal()`` in ``web/js/runtime-rpc.js`` and parsed here with
+    # ``json.loads``, so every value a header names is a string.
+    #
+    # The value type is ``Any`` rather than ``str`` because a row may also carry
+    # non-header keys the editor rides along on it — currently the ``__geo``
+    # sidecar (``web/js/row-geo.js``), an object, which is why the type cannot
+    # promise ``str`` for arbitrary keys. Nothing in the runtime reads those: every
+    # consumer projects a row through ``CSV_HEADERS`` and ignores the rest, which is
+    # what keeps the sidecar out of a codeplug.
+    Row = dict[str, Any]
+    Rows = list[Row]
 
 # Driver extras ride on channel rows under a key that is not a CSV header, the
 # same way repeater coordinates do in web/js/row-geo.js. Everything that
--- a/web/python/webchirp_bridge/channel_rows.py
+++ b/web/python/webchirp_bridge/channel_rows.py
@@ -10,7 +10,7 @@ under ``ROW_EXTRA_KEY`` and are applied to memories here as well.
 
 from __future__ import annotations
 
-from typing import Any, Optional, Sequence
+from typing import Any, TYPE_CHECKING
 
 from chirp import (
     chirp_common,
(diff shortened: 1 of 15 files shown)


jasiek:
@codex review

web/python/webchirp_bridge/column_metadata.py
```diff
@@ -31,113 +31,74 @@ def _radio_supports_dv(rf: Any) -> bool:
...
+
+def _formatted_enum_column(
+    editable: Any, values: Optional[Iterable[Any]], fmt: str, convert: Callable[[Any], Any]
+) -> dict[str, Any]:
+    """An enumerated column of numbers, spelled the way an exported CSV spells them."""
+    options = [fmt.format(convert(value)) for value in (values or [])]
+    return {"kind": "enum", "editable": bool(editable), "options": options}
+
+
+def _freq_column(editable: Any, rf: Any) -> dict[str, Any]:
```
jasiek: Why do we use Any here? Shouldn't this be a Radio?

web/python/webchirp_bridge/power_levels.py
```diff
@@ -79,6 +79,19 @@ def _power_levels_by_label(levels: Iterable[Any]) -> dict[str, Any]:
     return mapped
 
 
+def _level_map_for_radio(radio: Any, module_name: str, class_name: str) -> dict[str, Any]:
```
jasiek: Same here, shouldn't this be chirp_common.Radio?

web/python/webchirp_bridge/radio_memories.py
```diff
@@ -91,6 +92,27 @@ def _log_grouped_channel_failures(
         )
 
 
+def _read_memory(radio: chirp_common.Radio, number: int) -> chirp_common.Memory:
```
jasiek, line 97: So we use this type here, but not elsewhere - why?

web/python/webchirp_bridge/column_metadata.py
jasiek: It is not a Radio: rf is the RadioFeatures object from radio.get_features(), so it is now typed chirp_common.RadioFeatures, and editable is bool. Fixed in d154560.

web/python/webchirp_bridge/power_levels.py
jasiek: This one can legitimately be None: the upload preflight calls it before any radio instance exists, and the body already branches on that (levels = ... if radio else []). It is now Optional[chirp_common.Radio], and the return is dict[str, chirp_common.PowerLevel]. Fixed in d154560.

web/python/webchirp_bridge/radio_memories.py
jasiek: Fair catch. The annotation pass on #154 typed what had no annotation and left the Any that the older, already-annotated signatures used for radio_cls, rf and the settings objects, so the two styles ended up side by side. d154560 makes the package consistent: chirp_common.Radio / type[chirp_common.Radio] / RadioFeatures / PowerLevel everywhere a radio-shaped value is meant, and the chirp_settings classes (plus a SettingsContainer alias for RadioSettings-or-RadioSettingGroup) in the settings walkers. Any is left only where the value genuinely varies: JS proxies, JSON payloads. One caveat: pyright cannot resolve chirp (reportMissingImports is off), so these annotations are the documented contract rather than a checked one.

jasiek:
Wherever we import a type solely for the purpose of fulfilling a type annotation, use import gated by typing.TYPE_CHECKING to not introduce dependencies at runtime.

jasiek:
Done in f87c86c. Per module, every imported name referenced only inside annotations (computed from the AST, so nothing used at runtime moved) now sits under if TYPE_CHECKING: the typing helpers, the cross-module Row/Rows/ValidationIssue aliases, and the types module where it only served ModuleType annotations. isinstance targets, constructors and module-level aliases stay as normal imports. The SettingsContainer alias moved into the guard too, since only the checker reads it. Safe because every module already has from __future__ import annotations. The rule is now in CLAUDE.md so it applies to future Python changes as well.

web/python/webchirp_bridge/channel_rows.py
```diff
@@ -10,7 +10,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Optional, Sequence
+from typing import Any, TYPE_CHECKING
```
jasiek: Shouldn't Any be in the clause below?

web/python/webchirp_bridge/row_validation.py
```diff
@@ -11,31 +11,62 @@
 from __future__ import annotations
 
 import re
-from typing import Any, Literal, Optional
+from typing import Any, Literal, TYPE_CHECKING
```
jasiek: Same here - I think we should only import TYPE_CHECKING here.

web/python/webchirp_bridge/channel_rows.py
jasiek: Yes. It was kept at runtime only because Row = dict[str, Any] is an executed module-level statement; the alias itself is annotation-only, so in 54f9eb2 both Row and Rows moved into the guard alongside Any, and TYPE_CHECKING is the only typing import left at runtime.

web/python/webchirp_bridge/row_validation.py
jasiek: Same fix in 54f9eb2: ValidationIssue, ValidationMessage and RowChangeAction moved into the guard with Any and Literal, so the module imports only TYPE_CHECKING at runtime. Every other module in the package was already in that state.

PR #157, opened 2026-09-08
Measure and publish test coverage for both the JavaScript and Python halves

jasiek:
Adds coverage measurement for both languages, publishes it where GitHub can show it, and gates against regression. Nothing in `web/` changes: this is test and CI infrastructure.

## Where coverage stands

| scope | lines | branches | functions | files |
| --- | ---: | ---: | ---: | ---: |
| JavaScript (`web/**`) | 83.13% | 80.94% | 80.21% | 44 |
| Python (`webchirp_bridge`) | 85.47% | — | — | 14 |

## What changed

**Every shipped module is now loaded by a test.** V8 coverage only reports files a test actually imported, and `--test-coverage-include` does not force-load anything, so `web/app.js`, `web/js/runtime-rpc.js`, `web/js/tooltip.js` and `web/js/version-info.js` were *absent* from the report rather than listed at 0% — 805 lines outside the denominator. `scripts/test-module-loading.mjs` walks `web/` and imports everything it finds, so a new module is covered the day it lands. The headline figure drops from 85.12% to 83.13% as a result; the lower number is the honest one.

`runtime-rpc.js` imports Pyodide from an absolute jsDelivr URL — the only specifier that works in a browser with no bundler, and one Node cannot resolve at all. That single line, not any property of the module, is why 600 lines of core RPC logic had no test. A `module.register()` resolve hook maps the URL onto the pinned `pyodide` package in `node_modules`.

**Python is measured inside Pyodide.** Pyodide ships coverage.py and its C tracer works under WASM, so this measures the real runtime rather than approximating it. Three constraints, all documented in FINDINGS.md: tracing must start between `loadPyodide()` and `seedPyodideRuntime()`; it must use `include=` and not `source=`, because coverage starts before the seed writes those files and `source=` treats a missing directory as a module name and then reports a silent zero; and `runtime_bridge.py` is not instrumentable at all, since it is the one runtime file executed as a string, so CPython compiles it as `<exec>`. The report says so rather than dropping it.

**`npm run coverage`** runs the suite once and writes `coverage/js.lcov`, `coverage/python.lcov`, `summary.json` and `summary.md`.

## Reporting into a PR

A new `Coverage` workflow runs on pull requests, which nothing did before — `pages.yml` tests on master, which is too late to learn a branch is broken.

- **A table in the job summary**, plus lcov and summary.json as a downloadable artifact.
- **Patch coverage**: of the lines a branch changes, how many the suite executes. Forty untested new lines move a 13,000-line project total by a third of a percent, so only the patch figure can answer \"did this branch test what it added\".
- **Annotations on the diff** for uncovered changed lines, next to the code that caused them.
- **One comment per PR**, edited in place on each push, found by a hidden marker so it can never edit somebody else's.

Three GitHub limits are handled rather than discovered later: ten annotations render per step, a comment body is capped at 65,536 characters, and a fork PR gets a read-only token and cannot be commented on at all — that step is skipped by condition, with the job summary and artifact still carrying everything.

## Tracking, without a third-party service

Each master run uploads its `summary.json`; a PR downloads the last successful one and reports the change against it. `coverage-floors.json` is committed, so a branch that improves coverage raises the floor in the same commit and `git log` over that one file is the trend line.

## Notes for review

- **No new dependencies.** No additions to `package.json` dependencies or devDependencies. coverage.py comes from Pyodide's own package set at test time and is never shipped.
- **The diff base is the base branch, not `pull_request.base.sha`.** Your own `pr-refs-go-stale` finding says GitHub pins that SHA at PR creation and never advances it, so a week-old branch would be blamed for every master commit since. Three-dot diff against the ref lets git recompute the merge base. That needs `fetch-depth: 0` on checkout.
- **Patch coverage is reported, not gated.** A branch can legitimately add a line no unit test reaches, and a gate that cries wolf gets ignored. The project floors do gate.
- **Floors carry a 0.5-point tolerance.** Coverage is not deterministic: three runs of one commit gave JS branches 81.04%, 80.94%, 80.94%, because `test-driver-import-race.mjs` races imports on purpose. A floor set at the last measurement fails on an unchanged branch. Do not hand-edit `coverage-floors.json` up to a measured value; use `npm run coverage -- --update-floors`.
- **Two caveats on the numbers**, both in FINDINGS.md. Node counts every physical line including comments, so the JS line figure is not the statement coverage coverage.py reports and the two are not comparable to each other; branch and function percentages have no such caveat. Changed comment lines are therefore filtered out of the JS patch figure by a heuristic that can only remove lines, never invent them.
- **master was merged in and coverage re-measured** before opening this, since the branch predates a substantial rework of the Python bridge and CI checks out the merge. Floors hold.

`scripts/test-coverage-patch.mjs` adds 14 tests over the diff arithmetic that is easy to get wrong: a hunk header with no count, a pure deletion, a rename, a changed comment, range collapsing, and both truncation limits. Two gaps in the shared fake DOM surfaced and are fixed: unquoted attribute selectors (`input[type=checkbox]`, which `serial-test-page.js` ships) and a missing `getElementById`.

Full suite green: 1,152 tests, pyright clean.


jasiek:
@codex review

jasiek:
Actioned the Codex review — all six findings were real and are fixed in 04a5dac.

**Capability stubs made absent, not undefined** (P2 ×2). The sharpest one. The module-loading test set `navigator.serial` and `navigator.usb` to `undefined` to simulate an unsupported browser, but `web/js/serial.js:40-45` and `web/js/serial-test-page.js:64-69` detect support with the `in` operator, so a key present and undefined reads as *supported* — the exact opposite of the comment sitting above it. Both keys are now absent.

**`build:dist` added before the tests** (P2). Correct catch: `pages.yml` runs it first because `build:catalog` regenerates `web/radio-catalog.json` from the submodule pin and three tests read that file, so a pin or catalog change could pass here and only be validated after merge — which is the thing this workflow exists to prevent. Measured at 3.4s, so there was no trade-off to weigh.

**`coverage/` created before the comment is written** (P2). Right, and there was a second-order version of it: `coverage:patch` and `coverage:report` also assume the directory exists and would have failed on top of the original error, reporting three red steps for one cause. Both now say why there is no data instead.

**Dependabot skipped alongside forks** (P2). Not currently reachable — there is no `.github/dependabot.yml` — but repository-level security updates can create PRs without one, so the guard is worth having before it bites.

**Annotation-only import gated** (P1). Correct, and the fix improves on the suggestion: rather than keeping `Any` behind `TYPE_CHECKING`, the variable is now annotated `coverage.Coverage | None`, which is both accurate and a better match for the convention. Verified the tracer still collects (830/1383 statements from one test file) — the annotation must not be evaluated at runtime, since `coverage` only exists after `loadPackage()`.

**Canonical paths — fixed mine, declining the sweep** (P1). Both comments named are mine and are corrected. Worth flagging for whoever reads this though: **14 further bare-filename references predate this branch** — in `build-catalog.mjs`, `retain-deployed-assets.mjs`, `test-build-dist.mjs`, `test-debug-log.mjs`, `test-driver-import-race.mjs`, `test-runtime-bootstrap.mjs`, `test-serial-test-page.mjs` and `test-ui-radio-loading.mjs`. The reason none were caught is that the enforcing test scans `sourceFiles(webDir)` only, which also matches the rule's stated rationale: `build-dist.mjs` rewrites references textually in the files it builds, and it does not build `scripts/`. Sweeping those and extending the test to cover `scripts/` is a separate change; doing it here would bury this diff in unrelated edits.

Full suite green at 1,163 tests, pyright clean.

PR #162, opened 2026-09-09
Edit a channel's driver extra attributes from a modal behind an Extra column

jasiek:
Adds an editor for a channel's driver-specific extra attributes (CHIRP's `Memory.extra`), reached from a button in an **Extra** column that sits last in the channel grid, after Comment.

## Why

`Memory.extra` is where a driver keeps everything the 21 CSV columns have no room for — Busy Channel Lockout, PTT-ID, signalling code, scramble, compander. The runtime already carried those values to and from the editor on the row itself, under the `__extra` sidecar, but nothing could change them: the sidecar is names to bare values, and "1", "on" and `3` say nothing about what else the driver would accept.

## What changed

**Runtime** — new `web/python/webchirp_bridge/channel_extra.py` takes over the whole extras concern from `channel_rows.py`: the sidecar key, the read side, the replay, and now the schema. `get_channel_extra(module, class, location)` reads the memory the row occupies and serializes the driver's own setting objects with the radio-wide settings panel's `_serialize_setting_value`, so a driver extra describes itself exactly as a radio-wide setting of the same type does. Fields the sidecar could not carry back — multi-value settings, non-primitive values — are dropped rather than offered. New `getChannelExtra` RPC.

**UI** — `web/js/ui/channel-extra.js` is the modal; `web/js/row-extra.js` is the browser side of the sidecar key, mirroring `row-geo.js`. The modal opens before the runtime is asked anything (that first call can be the one that boots Pyodide), fetches the schema per open, and overlays the row's stored values on it — so a channel moved to another slot shows *its own* settings rather than the ones sitting in the slot. Immutable fields render disabled and are never written back. Saving merges the mutable values into the row sidecar; the upload path was already the thing that writes them to the radio.

## Two things a reviewer should look at

**1. An upload shortcut had to be reopened.** `_prepare_row_change()` skips a row that is column-for-column equal to its driver memory — that is what keeps an untouched codeplug from being rewritten, and it is *exactly* what an extras-only edit looks like, since the editor writes to the sidecar and to nothing else. The skip branch in `_apply_rows_to_radio_instance` now replays the sidecar against the memory it has already read, writing only when a value actually differs, so an untouched row still costs nothing. `tests/channels/channel-extra-fields.mjs` covers this and fails on the old skip path.

**2. The column is synthetic, and pinned.** It is appended to the *rendered* columns and deliberately **not** added to `state.currentHeaders`: every serializer — TSV clipboard, CSV export, the upload payload — projects rows through that list, and a header no CSV knows would have to be filtered back out at each one. It appears only once a loaded row carries extras, which is the same condition under which a schema can be read at all.

The first version of the column looked broken in exactly this way: reported as "I drag an image onto the window and I don't see any extra attributes". The column was there and populated — the drop path and the Load Binary button share `loadCodeplugFile` — but Extra is the eighteenth of eighteen columns, and the grid is 1185px wide in a ~620px viewport, so it sat past the right edge with TStep the last thing on screen. It is now sticky at `right: 0` of the grid's scroll area: still last, but in view at every horizontal scroll position. Same two constraints as the existing sticky header — an opaque background, and a `box-shadow` standing in for the left border that `border-collapse` hands to the table and that therefore scrolls away with it.

## Verification

Driven in a real browser against the `Baofeng_BF-888` upstream test image, via a genuine drop event: the Extra column and header pin to the right edge at scroll 0, read normally when scrolled fully right, and clicking a pinned button on row 4 opens "Extra settings for channel 4" showing the h777 driver's Busy Channel Lockout with its own `set_doc()` explanation, plus Beat Shift(scramble). Toggling and saving writes to the row.

All 1252 tests pass. Coverage floors met: JS 84.16% lines / 80.83% branches / 81.69% functions, Python 85.61% lines. The stylesheet test was verified by removing `right: 0` and watching it fail, rather than by watching it pass.

## New dependencies

None.

## Known limits

- A driver that exposes no per-channel extras gets no column, and the app currently says nothing about why. Same for a grid loaded from CSV: no memory was decoded, so no schema is readable.
- No bulk edit — one channel at a time, where desktop CHIRP can edit an extra down a whole column.
- The schema is fetched per open rather than cached per driver. Correct if a driver varies extras by slot, but it is a Pyodide round trip per click.


jasiek:
@codex review

jasiek:
Worked through the review in 05fd976 — nine findings taken as reported, one taken somewhere else. Detail per finding:

**Initialize sidecar-free rows from actual defaults (P1)** — correct, and the most serious of the ten. A channel with no sidecar opened on the values in the memory its Location points at, and the save wrote all of them back, so changing one setting on a replacement channel handed it every setting of the channel it replaced. The save now records **only fields the user actually changed**; an untouched form writes no sidecar at all, since its absence is what lets the driver apply its own defaults; and the modal now says whose values it is showing when a channel carries none of its own. A test of mine asserted the old behaviour and has been corrected.

**Fail the operation when an extra value is rejected (P1)** — agreed on the defect, taken at a different layer. Raising in `_apply_row_extras_to_memory` would fire with a clone in progress and memories already written, leaving a half-programmed radio. The refusal is now reported by the **upload preflight** (`_row_extra_findings`, against the Extra column), so it stops the operation while it is still a highlighted cell; the write path logs the driver's full traceback and continues, which also settles the AGENTS.md traceback point.

**Validate memories after attaching edited extras (P2)** — correct, and it predates this branch: a row equal to its memory returns from `_prepare_row_change` before `validate_memory`. The preflight now runs the driver's validation on the skip path with the sidecar applied. Confirmed the three drivers you named are the ones that read `mem.extra` in `validate_memory` (`hf90`, `ft450d`, `ar8200`), measured against the pinned CHIRP revision.

**Parse the whole integer input (P2)** — correct. `Number()` plus `Number.isInteger`, with a test that `1e1` reads as 10.

**Reject saves after the row set is replaced (P2)** — correct. The save now verifies the row is still in `state.currentRows`, tested by loading a second image with the modal open.

**Move focus into and back out of the modal (P2)** — correct. The grid hands its button to the editor (a click does not focus a button on every platform, so reading `activeElement` would not do), focus moves to Cancel the moment the dialog opens rather than when fields render, and returns to the button on every close path including Escape.

**Keep large extras forms inside the viewport (P2)** — correct. The card caps at `calc(100dvh - 2rem)` and the field grid scrolls. Checked with an 18-field form at 375x812: card bottom 796px, Save and Cancel reachable.

**Restore the single-column mobile layout (P2)** — correct, and the specificity analysis was right. Added an id-qualified override in the same media query; verified the grid computes to one column at 375px.

**Use the canonical path in the stylesheet comment (P1)** — correct, fixed. Two pre-existing partial paths for `repeater-map.js` (`web/styles.css` and `web/index.html`) have the same problem but predate this branch, so they are left for a separate change.

**Name image export explicitly in the save status (P2)** — correct, fixed; the status now names image export and says CSV export does not carry these.

One thing the review did not raise that is worth recording: `RadioSettingValueBoolean.set_value` is `bool(value)` and so refuses nothing, which is why the rejected-value case is pinned on a UV-5R (PTT ID is a `RadioSettingValueList`) rather than on `h777`, whose channel extras are both booleans.

All 1261 tests pass.

PR #166, opened 2026-09-13
Let a Frequency cell be cleared instead of snapping back

### LINKED ISSUE #93: A Frequency cell can never be cleared: empty input snaps back to the previous value
## Summary

`parseFreqToHz("")` returns 0. `normalizeValue` then runs `inAnyBand(0, bands)`, which is false for any radio with `valid_bands`, and returns `previous`. `commitCellValue` writes that back into the input. Since blank Frequency is the documented way to erase a memory on upload (`_apply_rows_to_radio_instance` calls `erase_memory` for it), there is no way to empty a single channel from the grid short of Remove, which triggers the renumbering bug in #73.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/channel-values.js#L5-L18
- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/channel-values.js#L67-L77
- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/channel-table.js#L930-L936

Yes. Parser and normalizer re-read. Not reproduced at runtime.

## Reproduction

1. Load any codeplug.
2. Select a Frequency cell, delete its contents, press Tab.

Expected: cell stays empty. Actual: the old frequency reappears.

## Suggested fix

Treat an empty string as a valid "erase" value for Frequency in `normalizeValue`.

## Priority

Medium.

---

## Triage 2026-09-10 (against e0c906f)

Supersedes the 2026-09-07 pass. Still valid — and notably **not** fixed by `3b6d2f4`, which added exactly the mechanism this needs.

`web/js/ui/channel-values.js:5-11` — empty still parses to `0`, not `null`. The normalizer is now named `normalizeCellValue` and returns a `{value, accepted}` pair (`channel-values.js:87-96`):

```js
if (meta.kind === "freq") {
  const hz = parseFreqToHz(v);
  if (hz === null) { return rejected(previous); }
  const shouldCheckBands = column !== "Offset";
  if (shouldCheckBands && !inAnyBand(hz, meta.bands || [])) { return rejected(previous); }
```

`0` is not `null`, so it reaches the band check, and `inAnyBand(0, bands)` is false for any driver with `valid_bands` (`:20-26` returns true only for an empty band list).

**Why `3b6d2f4` did not fix this:** the grid still calls the value-only wrapper `normalizeValue` (`channel-values.js:121-123`), which discards the new `accepted` flag. `channel-table.js:1221-1226`:

```js
const next = normalizeValue(column, editor.value, meta, row[column]);
row[column] = next;
editor.value = next;
```

Reached from the `focusout` handler at `:1267-1275`. So the rejection signal that #104's fix introduced exists but is thrown away on this path — reading `accepted` here is most of the fix.

The erase contract survives the package split: `web/python/webchirp_bridge/row_validation.py:255-267` ("Empty frequency means erase") feeds `radio_memories.py:238-239` (`if action == "erase": _erase_memory(radio, number)`).

Workaround unchanged: Remove (toolbar `#channel-remove`, `web/index.html:217` → `channel-table.js:475-505`) deletes the row, and an omitted memory number is erased on upload (`radio_memories.py:246-269`) — so the on-radio effect matches, but the slot disappears from the grid rather than being blanked.

Shares a root cause with #94. The third member of that cluster, #104, was fixed on 2026-09-08.

jasiek:
Fixes #93.

## What was wrong

`parseFreqToHz("")` returned `0`, so a cleared Frequency cell reached the band check as a 0 Hz frequency, which is in no driver's `valid_bands`. `normalizeCellValue` rejected the write and `commitCellValue` wrote the previous frequency back into the input on blur. Since an empty Frequency is how the runtime is told to erase a memory (`_prepare_row_change` in `web/python/webchirp_bridge/row_validation.py`), there was no way to blank one channel from the grid.

## The fix

- `web/js/ui/channel-values.js`: the `freq` branch returns `{ value: "", accepted: true }` for blank (or whitespace-only) input, before the parse and band check. Applies to Offset too, where blank simply means no offset.
- `parseFreqToHz("")` now returns `null` rather than `0`, so blank is not mistaken for a zero frequency anywhere else; the blank meaning lives in the one place that has it.

## Tests

`tests/channels/channel-values.mjs` gains a case covering blank Frequency and Offset, the `normalizeValue` wrapper, and the new `parseFreqToHz` blank result. `npm test` passes (syntax, pyright, and all four suites).

## Dependencies

None added.


jasiek:
@codex review

Issue #146, opened 2026-09-07
Feature: Bulk-edit channels


jasiek:
Expose a feature which, when a group of channels is selected allows you to edit their attributes, including extra settings.

- Select one or more channels
- Click on "edit in bulk"
- Modal appears which contains a list of attributes (what is in columns) and extra settings
- Whatever state is set in this modal gets reflected in the selected channels

PR #169, opened 2026-09-13
Bulk-edit channels (claude-opus-5, effort low)

jasiek:
Selecting one or more channels in the grid and pressing the new "Edit in bulk"
toolbar button opens a modal listing every editable grid column for the selected
radio, plus the driver's per-channel extra settings (CHIRP's `Memory.extra`, the
same schema the per-channel Extra editor renders).

Each field has an opt-in checkbox beside it. Applying writes only the ticked
fields, and writes them to every selected channel; anything left unticked keeps
whatever value each channel already had. Location is never offered (it is the
memory slot, per channel by definition), nor are columns the driver marks
read-only or extras it marks immutable.

Column values go through the grid's own validating writer, so a value the driver
rejects (an out-of-band frequency, an off-list enum) leaves that channel
unchanged and is counted in the status line. Extra settings are merged into the
row sidecar, so unrelated stored extras survive. If the driver's extras cannot be
read, the modal stays usable for the column fields and says so.

New dependencies: None.

Model: claude-opus-5. Effort level: low.

Refs #146


jasiek:
Mostly functional

PR #172, opened 2026-09-13
Keep a half-typed cell value across a scroll or resize re-render

### LINKED ISSUE #94: Half-typed cell values are committed and reverted on every grid scroll or resize
## Summary

`renderRowWindow` runs on each `scroll` event and ResizeObserver tick and calls `captureFocusedCell()`, which commits the focused input through `commitCellValue`. For Frequency, a partial value like "146." (regex fails) or "14" (out of band) normalises to the previous value and `editor.value = next` overwrites what the user typed. A related handoff issue: `restoreFocusedCell` focuses the new element while the old element has already been rebound to another row, so the synchronous `focusout` commit runs against that other row and can truncate a Name the user never touched.

## Where

- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/channel-table.js#L789-L819
- https://github.com/jasiek/webchirp/blob/b4843148441719545b44a1fb8e92567159e9c4e6/web/js/ui/channel-table.js#L968-L976

## Reproduction

1. Load a codeplug with more rows than fit on screen.
2. Click a Frequency cell and type "146." without leaving the cell.
3. Nudge the mouse wheel over the grid.

Expected: typed text is preserved. Actual: the cell snaps back to the old value.

## Suggested fix

Stash the raw editor text (not a normalised commit) when re-rendering, and restore it verbatim; commit only on blur/Enter.

## Priority

Medium.

---

## Triage 2026-09-10 (against e0c906f)

Supersedes the 2026-09-07 pass. All three sub-claims re-verified; line numbers all moved (`renderRowWindow` 985→1125, `captureFocusedCell` 952→1092, `cellElement` 829→951, `commitCellValue` 1093→1221).

**1. Scroll/resize revert — still valid.** `renderRowWindow` (`channel-table.js:1125-1148`) still opens with `const focused = captureFocusedCell();`, and `captureFocusedCell` (`:1092-1109`) still commits unconditionally via `commitCellValue(cell, active)` at `:1108`. Triggers unchanged at `:1277-1284` (`scroll` listener plus `ResizeObserver`). `"146."` fails `/^\d+(\.\d+)?$/` (`channel-values.js:12-14`) → `rejected(previous)` → `editor.value = next` clobbers the caret text. The double-commit amplifier also survives: `:1138-1147` still recurses once when `measureRowHeight()` reports a new height.

**2. The focus-handoff claim stays retracted.** `captureFocusedCell()` still runs at `:1130`, before the `bindRowElement` loop at `:1133-1135`; `cellReferenceFor` (`:1206-1213`) still reads `td.parentNode.dataset.rowIdx` at event time; `bindCellEditor` (`:871-905`) sets `editor.value` to the new row's stored value before the handoff `focusout` fires. Still a no-op.

**3. Caret left in a different channel — still valid.** `channel-table.js:951-957`:

```js
function cellElement(rowIdx, column) {
  const tr = rowElements[rowIdx - windowStart];
  if (!tr || tr.dataset.rowIdx !== String(rowIdx)) { return null; }
```

`restoreFocusedCell` (`:1114-1118`) bails on `null`. Row elements are recycled in place — `syncRowElementCount` rebuilds `tableBody` only when the row *count* changes (`:1050-1066`) — so on a pure scroll the focused `<input>` keeps focus while `bindCellEditor` repoints it at another channel, and subsequent keystrokes edit that channel. This still deserves its own issue.

`f816223` (sticky headers) and `6c5784b` (field-sizing) touched only styling and column widths; `renderRowWindow` and `measureRowHeight` are behaviourally unchanged.

**Incidental, found while checking the above and not part of this issue:** `visibleRowRange` (`channel-table.js:1039-1041`) still says "The header scrolls with the rows, so it offsets every row's position" and subtracts `headerHeight` from `scrollTop`, but the header is `position: sticky` since `f816223` (`web/styles.css:1581`). The subtraction biases the window up by about one row; `OVERSCAN_ROWS = 8` (`:44`) absorbs it, so it is a stale comment and an asymmetric overscan rather than a visible bug.

jasiek:
Fixes #94.

## What was wrong

`renderRowWindow` opened with `captureFocusedCell()`, which committed the focused editor unconditionally. It runs on every `scroll` event and every ResizeObserver call, so a value still being typed was normalized mid-edit: `"146."` fails the frequency parse and `"14"` is out of band, both of which fall back to the previous value, and `commitCellValue` wrote that back into the input over the caret. Nudging the wheel while typing a frequency reverted the cell.

## The fix

`web/js/ui/channel-table.js`:

- `captureFocusedCell` no longer commits. It carries the editor's raw text as `draft` alongside the cell reference and selection.
- `restoreFocusedCell` writes that draft back over the stored value `bindRowElement` just put in the recycled element, and re-anchors the selection — rewriting `value` collapses the caret to the end, so the selection is restored whenever either the text or the focus moved.
- A draft whose channel fell out of the rendered window has no element to return to, so it is committed there, exactly as a blur would. Nothing typed is silently dropped.
- `commitCellValue` is split: `commitRawValue` normalizes a string into the row and returns what was stored, so the row write can happen with no element to write back to, and it leaves the editor alone when the row is gone.

Committing still happens where it always did: blur (`focusout`), Enter, select `change`, and toolbar clicks, which blur the editor before they fire.

Out of scope, noted in the issue's triage: on a pure scroll the focused `<input>` keeps focus while being repointed at another channel, so subsequent keystrokes edit that channel. That is a separate defect and wants its own issue.

## Tests

New `tests/channels/ui-cell-draft.mjs`, driving the real UI controller against the fake DOM:

- a half-typed `"146."` survives a scroll re-render
- an out-of-band partial `"14"` is not reverted mid-typing
- leaving the cell still commits, and still rejects the invalid value
- an accepted edit committed on blur outlives the next re-render

The first two fail on the pre-fix code and pass after. `npm test` passes (syntax, pyright, and all four suites).

## Dependencies

None added.


jasiek:
@codex review

PR #174, opened 2026-09-13
Static per-model pages for the 55 Baofeng radios

jasiek:
## Why

The app is a single shell that says nothing until Pyodide has booted, so a crawler sees roughly 80 words of text and the site cannot answer the searches it exists to serve — "baofeng uv-5r programming software", "bf-888s driver", "program do baofeng uv 5r". The word "Baofeng" did not appear in `web/index.html` at all.

## What this adds

**`radio-features.json`** — what each of the 554 catalogued radios can do, read from the driver's own `RadioFeatures` through the same cached-or-blank instance the channel grid uses. Written by the existing catalog build, where every driver is already imported. It lives outside `web/` on purpose: the page generator reads it at build time and no browser ever fetches it.

**`web/radios/`** — one page per Baofeng radio plus a vendor index, generated by `scripts/build-model-pages.mjs`. Every fact on a page comes from the driver, which is what keeps 55 pages from being 55 copies of one template:

- UV-5R: 128 channels, 7-character names, High 4.0W / Low 1.0W, three bands
- BF-888: 16 channels, 6-character names, UHF only

Rebadges come from the driver's `ALIASES`, so a Retevis RT5R owner finds the UV-5R page and is told it covers them. Each page carries a `FAQPage` JSON-LD block and links its siblings.

**`?radio=<catalog key>`** — each page's call to action opens the app with its radio already selected. The parameter outranks the last-radio cookie for that visit and falls through to it when the key is unknown, which is the expected case after a CHIRP bump.

**`sitemap.xml` and `robots.txt`** — written from the same page list. Neither existed before.

Widening past Baofeng is one line: the `VENDORS` set in `scripts/build-model-pages.mjs`. Held back deliberately so the template can be measured before it is multiplied by ten.

## Two bugs found on the way

`icw32.ICW32ARadio` reads its band limits out of the codeplug, so a blank instance reports 0 Hz to 0 Hz — its page would have advertised that. Degenerate ranges are now dropped, and `FINDINGS.md`'s `blank-instances-misreport-state` entry is widened with the band case.

`radio-features.json` differed between two builds of an unchanged tree, in 14 of 554 radios, with nothing behind the diff: several drivers build their capability lists as `list(set(...))` (`ft817.py:443`, `id31.py:212`), and a Python set of strings iterates in a different order in every process. The lists are sorted at the bridge boundary now, and two consecutive builds were verified byte-identical. Recorded as `radio-features-are-unordered-at-the-source`.

## Not committed

`web/radios/`, `sitemap.xml` and `robots.txt` are generated at build time and gitignored. Both workflows already run `build:dist` before `npm test`, and `build:dist` runs `build:pages`, so the committed copies were never the ones tested or deployed — and the diff they produced would grow to a page per radio once `VENDORS` widens. `npm run dev` builds them first so the local server still serves them.

The test runs the real generator against a throwaway tree with the actual catalog copied in, so it builds what it asserts on rather than depending on build order.

## Dependencies

None added.

## Open question

CHIRP calls the radio **BF-888**; four of the target search terms say **BF-888S**. I used CHIRP's name rather than inventing a model designation. If the page should match the search, the honest fix is a sourced line in the alias section — say the word and I will add it.

## Validation

`npm test` green: 632 / 161 / 562 / 35, zero failures. `npm run build:dist` produces all 56 pages with asset references rewritten correctly from the subdirectory.


jasiek:
@claude review

jasiek:
@codex review

PR #179, opened 2026-09-15
Retry transient failures when seeding the runtime sources

jasiek:
## Problem

A bare page load can file a Sentry "runtime crash" when the Pyodide bootstrap fetches a core CHIRP file or bridge module and the request fails at the network layer. `seedPyodideRuntime` seeds ~30 files in a `Promise.all`, so **one rejected request aborts the whole boot**: the visitor gets "Initialization failed; sidebar controls remain disabled" and a reload is the only way forward.

The original report's stack pinpoints it — the top frame is Sentry's instrumented `fetch` (`@.../@sentry/core.../+esm`), not `fetchText`'s own `!res.ok` throw (which would name `web/js/python-sources.mjs`). That is a transport rejection: offline, DNS/TLS, a blocking proxy, or a CDN 429/5xx — not a 404. Retrying is safe because every one of these requests is an idempotent GET for a pinned asset.

A second, separate failure mode is a connection the host accepts and then never answers: `fetch()` neither resolves nor rejects for the browser's ~300 s default, so the boot sits there with no way to tell a slow CDN from a dead one.

## Change

- **Bounded retry** (`web/js/python-sources.mjs`): up to `PYTHON_SOURCE_MAX_ATTEMPTS` (3) attempts with jittered exponential backoff. The 30-odd seeding requests run at once, so jitter matters or a rate-limited CDN stays rate-limited.
- **Per-attempt deadline** (`PYTHON_SOURCE_TIMEOUT_MS`, 15 s): an `AbortController` + `setTimeout`, with the body read inside the attempt so a response that stalls mid-body is bounded too. This mirrors the reasoning in `web/js/request-timeout.js`.
- **Transient-only**: transport rejections, timeouts, and `408`/`429`/`5xx` are retried; every other 4xx fails on the first attempt so a bad pin still fails loudly rather than after three waits.
- **Debug visibility**: each retry emits a `RETRY …` line to the debug panel via an injected `onRetry` callback.
- **No change to the bootstrap classifiers**: internal retries happen before `ensure()` classifies the final failure, so a retry that succeeds produces no crash report and an exhausted one still reports exactly once through `createBootstrapCrashReporter`.
- The retry lives in `createBrowserCdnPythonSource` only; `createFilesystemPythonSource` (the Node harness) is untouched and gains no network dependency.
- Records the failure-mode distinction and the retry policy in `FINDINGS.md`.

Not covered here (deliberate): `loadPyodide` is a separate module-level CDN import with no retry, and there is no user-facing retry on the init-failure path. Both are reasonable follow-ups, the second arguably higher-value for a persistent outage.

## Dependencies

None added.

## Tests

New `tests/channels/python-sources.mjs` (10 tests): transient retry until success and retry narration, retryable 5xx/429, permanent 404 fails after one attempt and keeps its message, attempt cap, `fetchChirpSource` CDN URL on failure, `listDriverModules` retry + parse, deadline abort with retry, connection abandoned on deadline, deterministic exponential backoff with jitter, and no sleep on success.

`npm test` (syntax, pyright, channels, webusb, settings, build) passes.


jasiek:
@codex review

PR #185, opened 2026-09-17
Stop a cancelled repeater query from hijacking the next directory

### LINKED ISSUE #180: Cancelling a repeater query leaves it in flight, so it inserts into and closes the next directory's modal
## Summary

An in-flight repeater-directory query is not cancelled by Cancel, and its completion hijacks whatever modal is open when it resolves. It inserts the old directory's repeaters and closes the new modal.

## Where

`web/js/ui/repeater-query.js`

- `queryInFlight` / `setQueryBusy` — lines 118–127
- `setModalOpen` — line 326
- `openModal` — line 354
- submit handler — around lines 470–515

## What happens

`queryInFlight` is one module-level flag for the single shared modal, and nothing clears it when the modal is closed or rebuilt:

- `setModalOpen(false)` only bumps `previewGeneration`; it does not touch `queryInFlight` or the submit button.
- `openModal(source)` rebuilds the form and re-shows the modal but never calls `setQueryBusy(false)`.

So this sequence is broken:

1. Open directory A and press **Query API**. `setQueryBusy(true)` disables the button and sets it to "Querying…" while `source.runQuery` is awaited. RSGB fans out over up to 24 locator squares, so this is seconds.
2. Press **Cancel** (or the backdrop). The modal hides, but the query promise is still pending and `queryInFlight` is still `true`.
3. Open directory B. `openModal` rebuilds B's fields, but the submit button is still disabled and still reads "Querying…" because nothing reset it. The user cannot run B's query at all.
4. When A's query finally resolves, its handler runs `recordFlow(...)`, then `setModalOpen(false)` — **closing B's modal** — after `ctx.table.insertRowsAtSelectionOrEnd(rows, ...)` has already inserted **A's** repeaters into the grid. Its `finally` then re-enables the button.

The `source` capture (added for the original in-flight guard) keeps the *query itself* correct, but nothing guards its *completion side effects* against a later modal.

## Why this is not #105 or #106

- #105 ("Repeater query form has no in-flight guard") was fixed by this flag; this is the residual state leak the flag introduces, a different scenario.
- #106 ("Opening a second repeater directory while the first is still loading its `/meta` options hijacks the modal") is about `openModal`'s `loadOptions` race, not an in-flight `runQuery`.

## Suggested fix

Make the busy state generation/token-scoped like `previewGeneration`: capture a token at submit and only let that submission call `setQueryBusy(false)`, `setModalOpen(false)` and insert rows if it is still current. Alternatively, have `openModal` and `setModalOpen(false)` reset or invalidate a submission belonging to a previous modal instance.

## Verification

- Not covered by `tests/channels/ui-repeater-modal.mjs`; its only in-flight query test is the double-submit case (#105), which never cancels and reopens.

jasiek:
Fixes #180.

## The bug

Cancel a repeater query while it is still running — an RSGB fan-out over up to 24 locator squares takes seconds — then open a different directory, and two things go wrong:

- The new directory's **Query API** button is still disabled and still reads "Querying...", so its query cannot be run at all.
- When the first query resolves it inserts the **previous** directory's repeaters into the grid and then closes the modal that replaced it.

The cause is that queryInFlight was a single module-level boolean for the one shared modal, and nothing invalidated it: the close branch of setModalOpen bumped only previewGeneration, and openModal rebuilt the form without resetting the button.

## The fix

The busy state is now generation-scoped, the same shape as previewGeneration and openGeneration beside it.

- A querySubmission token is bumped by abandonQuery(), called when the modal closes and when openModal rebuilds it for another directory. It also frees the submit button there, because the abandoned submission's own release is guarded and may arrive seconds later, or never.
- A submission captures the token and only closes the modal, records its flow outcome or releases the button while it still matches.
- The rows are inserted **inside** source.runQuery, before it resolves, so a check after the await in the caller would be too late. runQuery now takes an isCurrent() test as its second argument and asks it immediately before ctx.table.insertRowsAtSelectionOrEnd, in both the shared RXF directory source and the RSGB one.
- Debug visibility is kept: the request URL and the fetched/inserted counts are still logged, followed by a line saying the rows were discarded, and an error from an abandoned query goes to the debug panel with its stack instead of overwriting the new modal's status line.

## Verification

- New regression test in tests/channels/ui-repeater-modal.mjs: submit an RSGB query held open on a gate, cancel it, open przemienniki.net, then release. It asserts the button comes back, no rows are inserted and the new modal stays open. Verified to fail on master and pass with the fix.
- Full npm test green: syntax, pyright, channels (687), webusb (161), settings (564), build (44).

## Dependencies

None added.


jasiek:
@codex review this PR please.

